PR #55, opened 2021-06-22
Giver page

elisabethnetland:


RonHaland (dismissed):
Looks good to me.

Later, maybe we could look into moving texts into resource-files?
That way we can more easily change texts (and make translations).
I think this is sufficient for now.

GiEnJul/ClientApp/src/routes/registerGiver/LocationGiver.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Button, ButtonToolbar } from 'reactstrap';
+import { Route } from 'react-router-dom'
+import Dropdown from 'react-dropdown';
+import 'react-dropdown/style.css';
+
+type Props = {
+  nextStep: () => void,
+  prevStep?: () => void,
+  handleLocationChange: (newLocation: Object) => void,
+  values: { location: string | undefined; name: string | undefined; email: string | undefined; tlf: number | undefined; familiyType: string | undefined; }
```
joffe97, line 11: These could probably be shortened to e.g. `location?: string` instead of `location: string | undefined`. I think the result would be the same, but it would look a bit cleaner.

GiEnJul/ClientApp/src/routes/registerGiver/SignUp.tsx
```diff
@@ -0,0 +1,135 @@
...
+import LocationGiver from './LocationGiver';
+import SummaryRegistration from './SummaryRegistration';
+
+interface State {
+    step: number,
+    location: string | undefined,
+    name: string | undefined,
+    email: string | undefined,
+    tlf: number | undefined,
+    familiyType: string | undefined,
```
joffe97, line 14: These could probably be shortened to e.g. `location?: string` instead of `location: string | undefined`.

GiEnJul/ClientApp/src/routes/registerGiver/LocationGiver.tsx
elisabethnetland: Thanks! Made this a lot cleaner :D

joffe97 (approved):

PR #68, opened 2021-06-23
Added ci workflows

TheseusDeus:
To run a "dotnet test" command do I need to change the active directory first? 
"dotnet test" is not added to the main_backend.yml file yet

Map structure:
gi-en-jul
   → GiEnJul
     →→   *backend app
     →GiEnJul.Test

### DIFF
--- a/.github/workflows/main_backend.yml
+++ b/.github/workflows/main_backend.yml
@@ -4,12 +4,10 @@ on:
   push:
     branches: [ master ]
     paths: 'GiEnJul/**'
-    paths-ignore: 'GiEnJul/ClientApp/**'
       
   pull_request:
     branches: [ master ]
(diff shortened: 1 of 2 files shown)


.github/workflows/main_backend.yml
```diff
@@ -0,0 +1,32 @@
+name: Build backend service ASP.NET CORE 3.1 APP
+
+on:
+  push:
+    branches: [ master ]
+    paths: 'GiEnJul/**'
```
RonHaland, line 6: I think the backend needs to just exclude the **frontend folder.**

We might need to experiment a little with this.

TheseusDeus: So either exclude the frontend folder it in the main_backend.yml
or refactor so the folder appears at the root of the project

Issue #44, opened 2021-06-22
Backend - Add Api-function to recieve new registrations of Givers


Sommerli:
Create a GiverController. Implement a function to receive Givers and add them to Table Storage with GiverRepository.

This task is possible after #38 is completed.

PR #73, opened 2021-06-23
Connection with table storage

elisabethnetland:
Issue #44
Created controller in backend for giver as well as updated entities and models for giver in backend. 
Changed the fields of the state in registration giver in the frontend. 
Added a submit-function that sends a POST to tableStorage.

### DIFF
--- a/GiEnJul/ClientApp/src/routes/registerGiver/Confirmation.tsx
+++ b/GiEnJul/ClientApp/src/routes/registerGiver/Confirmation.tsx
@@ -7,10 +7,6 @@ type Props = {
 }
 
 const Confirmation: React.FC<Props> = ({ values }) => {
-    console.log(values.fullname)
-    console.log(values.email)
-
-
     return (
         <div>
             <h1>Takk for at du gir en jul {values.fullname}</h1>
--- a/GiEnJul/ClientApp/src/routes/registerGiver/Confirmation.tsx
+++ b/GiEnJul/ClientApp/src/routes/registerGiver/Confirmation.tsx
@@ -3,15 +3,18 @@ import { Route } from 'react-router-dom';
 import { Button } from 'reactstrap';
 
 type Props = {
-    values: { location?: string; fullname?: string; email?: string; phoneNumber?: number; maxRecievers?: number; familyType?: string; }
+    values: { location?: string; fullname?: string; email?: string; phoneNumber?: number; maxRecievers?: number; familyType?: string }
 }
 
-const Confirmation = () => {
+const Confirmation: React.FC<Props> = ({ values }) => {
+    console.log(values.fullname)
+    console.log(values.email)
+
 
     return (
         <div>
-            <h1>Takk for at du gir en jul</h1>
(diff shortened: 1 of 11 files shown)


GiEnJul/ClientApp/src/routes/registerGiver/Confirmation.tsx
```diff
@@ -2,11 +2,19 @@ import * as React from 'react';
...
+    console.log(values.fullname)
+    console.log(values.email)
+
+
     return (
         <div>
-            <h1>Takk for at du gir en jul</h1>
-            <p>Vi har nå registret deg som giver til årets Gi en Jul. vi har sendt en  oppsummering av din registrering i til Ola.Normann@gmail.com. Når det nermer seg jul vil vi sende deg mer informasjon. </p>
+            <h1>Takk for at du gir en jul {values.fullname}</h1>
+            <p>Vi har nå registret deg som giver til årets Gi en Jul og har sendt en  oppsummering av din registrering i til  { values.email}. Når det nermer seg jul vil vi sende deg mer informasjon. </p>
```
Sommerli, line 17: Just some typos. Text should probably be seperated to a .resx file at a later time, as Ronny has mentioned.
```suggestion
            <p>Vi har nå registrert deg som giver til årets Gi en Jul og har sendt en  oppsummering av din registrering i til  { values.email}. Når det nermer seg jul vil vi sende deg mer informasjon. </p>
```

elisabethnetland: Thanks, agreed! I have an idea of how to solve the editing of text via admin :D

GiEnJul/Controllers/GiverController.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        
+        // POST api/<GiverController>
+        [HttpPost]
+        public async Task PostAsync([FromBody] Models.Giver giver)
+        {   
+            _log.Debug("Adding giver object: {@giver}", giver);
+            var result = await _giverRepository.InsertOrReplaceAsync(giver);
+            _log.Debug("Succesfully added giver: {@0}", result);
```
Sommerli, line 34: I have two suggestions here:
Either remove unnecessary var
```suggestion
            _log.Debug("Adding giver object: {@giver}", giver);
            await _giverRepository.InsertOrReplaceAsync(giver);
            _log.Debug("Succesfully added giver: {@0}", result);
```
Or add return value. With inspiration from [documentation](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?view=aspnetcore-5.0&tabs=visual-studio#update-the-posttodoitem-create-method)
```suggestion
            //Change return type of function to Task<ActionResult<Giver>>
            _log.Debug("Adding giver object: {@giver}", giver);
            var result = await _giverRepository.InsertOrReplaceAsync(giver);
            _log.Debug("Succesfully added giver: {@0}", result);
            return CreatedAtAction(nameof(result), result);
```

Difference between the two is that frontend either receives an empty Http-Success response, or a Http-Success response with the added entity.

elisabethnetland: Done! Changed to last suggestion

TheseusDeus (approved):

PR #87, opened 2021-06-24
Implemented Connectionrepository

TheseusDeus:
Created connectionrepository with insert methods for giver and recipient models
Closes issue #76

### DIFF
--- a/GiEnJul/Infrastructure/AutoMapperProfile.cs
+++ b/GiEnJul/Infrastructure/AutoMapperProfile.cs
@@ -15,21 +15,23 @@ namespace GiEnJul.Infrastructure
             CreateMap<Models.Recipient, Entities.Connection>();
             CreateMap<Models.Giver, Entities.Connection>()
                 .ForMember(dest => dest.GiverEmail, opt => opt.MapFrom(src => src.Email))
-                .ForMember(dest => dest.SubmitterEmail, opt => opt.MapFrom(src => src.Receiver.ContactEmail))
                 .ForMember(dest => dest.GiverFullName, opt => opt.MapFrom(src => src.FullName))
                 .ForMember(dest => dest.GiverLocation, opt => opt.MapFrom(src => src.Location))
                 .ForMember(dest => dest.GiverPhoneNumber, opt => opt.MapFrom(src => src.PhoneNumber))
-                .ForMember(dest => dest.ReceiverLocation, opt => opt.MapFrom(src => src.Receiver.Location))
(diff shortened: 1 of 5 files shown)


GiEnJul/Models/Recipient.cs
```diff
@@ -5,6 +5,13 @@ namespace GiEnJul.Models
 {
     public class Recipient : TableEntity
     {
+
+        public Recipient(string location, string rowKey) : base(location, rowKey)
+        {
+            PartitionKey = $"loc_{location}";
+            RowKey = rowKey;
+        }
```
Sommerli, line 13: This constructor seems to be unused
```suggestion
```

TheseusDeus: That is true. Used it for testing and forgot to remove it

GiEnJul/Infrastructure/AutoMapperProfile.cs
```diff
@@ -11,6 +11,26 @@ public AutoMapperProfile()
...
+                .ForMember(dest => dest.PersonCount, opt => opt.MapFrom(src => src.Receiver.FamilyMembers.Count));
+            CreateMap<Models.Recipient, Entities.Connection>()
+                .ForMember(dest => dest.GiverEmail, opt => opt.MapFrom(src => src.Giver.Email))
+                .ForMember(dest => dest.SubmitterEmail, opt => opt.MapFrom(src => src.ContactEmail))
+                .ForMember(dest => dest.GiverFullName, opt => opt.MapFrom(src => src.Giver.FullName))
+                .ForMember(dest => dest.GiverLocation, opt => opt.MapFrom(src => src.Giver.Location))
+                .ForMember(dest => dest.GiverPhoneNumber, opt => opt.MapFrom(src => src.Giver.PhoneNumber))
+                .ForMember(dest => dest.ReceiverLocation, opt => opt.MapFrom(src => src.Location))
+                .ForMember(dest => dest.SubmitterFullName, opt => opt.MapFrom(src => src.ContactFullName))
+                .ForMember(dest => dest.PersonCount, opt => opt.MapFrom(src => src.FamilyMembers.Count));
```
joffe97, line 33: The mappings to SubmitterPhoneNumber is missing. It would be nice to have these sorted as well, so e.g. receiver mappings is next to other receiver mappings?

joffe97 (changes requested):

GiEnJul/Infrastructure/AutoMapperProfile.cs
TheseusDeus: Totally agree. This should be fixed now

joffe97 (approved):
Nice work!

PR #90, opened 2021-06-25
Fixed the recipient mappings

### LINKED ISSUE #89: Backend - Recipient mapping fix
Add mapping between recipient entity and model on the following variables:
* FullName
* Email
* PhoneNumber
This can be done by adding "Contact" in front of the variable name. E.g. "ContactFullName".

joffe97:
Fix #89 
Changed some recipient entities variables to match the recipient model variables, so the automapping works correct.


TheseusDeus (approved):
Sounds good. Will make the naming convention more clear as well

Issue #51, opened 2021-06-22
Frontend -Registartion Submit


TheseusDeus:
When issue #49, #46 ,#50 and bug #47 is fixed,
the submit form needs to be implemented

handleSubmit() function for the whole form
Should be easy to get the information right from state variables

TheseusDeus:
closed with #94

PR #94, opened 2021-06-25
implemented Institutions POST

TheseusDeus:
Naming is not superb, and some of this has to be refactored.
However to do that most of the excising "Person" interface and constant need to be  changed as well
closing issue #51

### DIFF
--- a/GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
+++ b/GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
@@ -7,7 +7,29 @@ import Location from './InstitutionLocation';
 import FormPerson from './FormPerson';
 import IFormPerson from './IFormPerson';
 import LOCATIONS from '../../common/constants/Locations';
+import Gender from '../../common/enums/Gender';
 
+
+type PersonType = {
+    Wish?: string,
+    Age?: number,
+    Gender?: Gender,
+    Family?: string,
+}
+
+type submittype = {
+    Dinner?: string,
+    Dessert?: string,
+    Note?: string,
+    Event?: string,
+    Location?: string,
+    ContactFullName?: string,
+    ContactEmail?: string,
+    ContactPhoneNumber?: string,
+    Institution?: string,
+    ReferenceId?: string,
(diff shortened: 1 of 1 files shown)


elisabethnetland (approved):

GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
```diff
@@ -61,8 +84,42 @@ const RegistrationForm = () => {
         if (e.target.value !== "annet") setDessertInput("");
     } 
 
+    const onSubmit1 = (e : React.FormEvent<HTMLFormElement>) => {
```
RonHaland, line 87: onSubmit1? are there multiple onSubmit-s here?
If there are they should probably have a name describing what the submit 😉

GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
```diff
@@ -61,8 +84,42 @@ const RegistrationForm = () => {
...
+            Event:"JUL2021",
+            Location:location,
+            ContactFullName:contactName,
+            ContactEmail:contactEmail,
+            ContactPhoneNumber: contactPhoneNumber,
+            Institution:"NAV",
+            ReferenceId:pid,
+            FamilyMembers: personsList
+        }
+        fetch('https://localhost:5001/api/recipent', {
```
RonHaland, line 112: This might need to be a config variable?
We can fix this later and potentially make a component that is responsible for communication with the back end.

GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
TheseusDeus: fair, that should be changed into something more understandable

GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
TheseusDeus: It should be a config variable yeah, but as you said that could be implemented later

RonHaland (approved):

PR #95, opened 2021-06-25
Recipient controller

### LINKED ISSUE #85: Backend - Create recipient controller

joffe97:
Fix #85


GiEnJul/Controllers/RecipientController.cs
```diff
@@ -0,0 +1,42 @@
...
+            var recipientId = recipient.RowKey;
+
+            foreach (var person in recipient.FamilyMembers)
+            {
+                person.PartitionKey = recipientId;
+                var presult = await _personRepository.InsertOrReplaceAsync(person);
+                _log.Debug("Succesfully added person: {@0}", presult);
+            }
+
+            var result = await _recipientRepository.InsertOrReplaceAsync(recipient);
```
TheseusDeus, line 37: if this fails, should all the persons inserted in the above function be deleted?

elisabethnetland (commented):
Looks great!

elisabethnetland (dismissed):
Looks nice!

GiEnJul/Controllers/RecipientController.cs
joffe97: Good point. I'll look into this!

GiEnJul/Controllers/RecipientController.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        [HttpPost]
+        public async Task<ActionResult<Entities.Recipient>> PostAsync([FromBody] Models.Recipient recipient)
+        {
+            _log.Debug("Adding recipient object: {@recipient}", recipient);
+            var recipientId = recipient.RowKey;
+            List<Entities.Person> addedPeople = new List<Entities.Person>();
+
+            for (var i = 0; i < recipient.FamilyMembers.Count; i++)
```
Sommerli, line 33: I suggest using `foreach(var person in recipient.FamilyMembers)` for readability

GiEnJul/Features/GenericRepository.cs
```diff
@@ -64,6 +67,28 @@ public async Task<T> DeleteAsync(string partitionKey, string rowKey)
...
+        {
+            List<T> deletedEntities = new List<T>();
+            foreach (var addedPerson in entities)
+            {
+                try
+                {
+                    var result = await DeleteAsync(addedPerson);
+                    deletedEntities.Add(result);
+                }
+                catch (Exception) { }
```
RonHaland, line 80: Log or keep these exceptions in a list to deal with later?
Currently we don't know what caused the delete to fail.

GiEnJul/Controllers/RecipientController.cs
```diff
@@ -0,0 +1,77 @@
...
+            _personRepository = personRepository;
+            _log = log;
+        }
+
+        [HttpPost]
+        public async Task<ActionResult<Entities.Recipient>> PostAsync([FromBody] Models.Recipient recipient)
+        {
+            _log.Debug("Adding recipient object: {@recipient}", recipient);
+            var recipientId = recipient.RowKey;
+            List<Entities.Person> addedPeople = new List<Entities.Person>();
```
RonHaland, line 31: this can be:
```
var addedPeople = new List<Entities.Person>();
```

GiEnJul/Controllers/RecipientController.cs
```diff
@@ -0,0 +1,52 @@
...
+            _recipientRepository = recipientRepository;
+            _personRepository = personRepository;
+            _log = log;
+        }
+
+        [HttpPost]
+        public async Task<ActionResult<Entities.Recipient>> PostAsync([FromBody] Models.Recipient recipient)
+        {
+            _log.Debug("Adding recipient object: {@recipient}", recipient);
+            foreach (var person in recipient.FamilyMembers) person.PartitionKey = recipient.RowKey;
```
joffe97, line 29: Is there a better way to map this?

Sommerli (changes requested):
Best practice would be to depend on interfaces and abstractions. I recommend changing from List to IList or even IEnumerable in Method parameters.  Look at table below for multiple cases.

https://stackoverflow.com/a/30646094
From this article https://www.claudiobernasconi.ch/2013/07/22/when-to-use-ienumerable-icollection-ilist-and-list/
image

GiEnJul/Controllers/RecipientController.cs
Sommerli: Could do:
`recipient.FamilyMembers.ForEach(person => person.PartitionKey = recipient.RowKey);`

GiEnJul/Features/GenericRepository.cs
joffe97: Fixed this by using batches.

GiEnJul/Controllers/RecipientController.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        [HttpPost]
+        public async Task<ActionResult<Entities.Recipient>> PostAsync([FromBody] Models.Recipient recipient)
+        {
+            _log.Debug("Adding recipient object: {@recipient}", recipient);
+            recipient.FamilyMembers.ForEach(person => person.PartitionKey = recipient.RowKey);
+            try
+            {
+                await _personRepository.InsertOrReplaceBatchAsync(recipient.FamilyMembers);
```
Sommerli, line 32: This operation might not need its "own" try catch, as the default error-handler will have almost equal outcome.

Sommerli (approved):

Issue #77, opened 2021-06-24
Connect response from backend when submitting a giver sign up


elisabethnetland:
**Use return from backend to show values for giver in confirmation**
In the backend in GiverController, function "PostAsync" we send a response of the entity that is saved in azure TableStorage to the frontend. Want to use this response to show the mail where giver will receive info in ClientApp/src/routes/registerGiver/Confirmation in frontend.

PR #116, opened 2021-06-28
added try catch for signup process

TheseusDeus:
added a simple try catch for signup process
will send user to the confirmation screen if response code is 201,
else just look at it as an error
closes issue #77 since the mail sent in the POST is now confirmed, and the state value mail can be used in the confirmation message

### DIFF
--- a/GiEnJul/ClientApp/src/routes/registerGiver/SummaryRegistration.tsx
+++ b/GiEnJul/ClientApp/src/routes/registerGiver/SummaryRegistration.tsx
@@ -3,24 +3,37 @@ import * as React from 'react';
 import { Button, Grid, TextField, FormControl} from '@material-ui/core';
 import useStyles from './Styles';
 
-
 type Props = {
     nextStep: () => void,
     prevStep: () => void,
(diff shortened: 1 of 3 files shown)


GiEnJul/ClientApp/src/routes/registerGiver/SummaryRegistration.tsx
```diff
@@ -3,25 +3,38 @@ import * as React from 'react';
...
+
+    const trigger = (b: boolean) => {
+        callingback(b);
+    }
 
-    const Submit = (e: any) => {
+    const Submit = async( e: any) => {
         e.preventDefault();
-        fetch('https://localhost:5001/api/giver', {
+        await fetch('https://localhost:5001/api/giver', {
```
Sommerli, line 23: Can change this to to keep the link relative
```suggestion
        await fetch('/api/giver', {
```

TheseusDeus: That would be better yes, will fix

joffe97 (approved):
Nice work!

Sommerli (approved):

PR #129, opened 2021-06-29
Added tests for recipient controller

joffe97:


GiEnJul/Controllers/RecipientController.cs
```diff
@@ -26,6 +26,11 @@ public RecipientController(IRecipientRepository recipientRepository, IPersonRepo
         public async Task<ActionResult<Entities.Recipient>> PostAsync([FromBody] Models.Recipient recipient)
         {
             _log.Debug("Adding recipient object: {@recipient}", recipient);
+            if (string.IsNullOrEmpty(recipient.Location)) {
+                var e = new Exception("Cannot add recipient with no location.");
+                _log.Error("Exception occurred while trying to add recipient: {@0}. \n{@1}", recipient, e);
+                throw e;
```
RonHaland, line 32:
```suggestion
                _log.Error("Exception occurred while trying to add recipient: {@0}", recipient);
                throw new ArgumentNullException("Location", "Cannot add recipient with no location");
```

Not sure if I used everything correctly here, maybe test this locally first

RonHaland (dismissed):

GiEnJul/Controllers/RecipientController.cs
```diff
@@ -26,6 +26,11 @@ public RecipientController(IRecipientRepository recipientRepository, IPersonRepo
         public async Task<ActionResult<Entities.Recipient>> PostAsync([FromBody] Models.Recipient recipient)
         {
             _log.Debug("Adding recipient object: {@recipient}", recipient);
+            if (string.IsNullOrEmpty(recipient.Location)) {
```
Sommerli, line 29: `if (string.IsNullOrEmpty(recipient.Location))`

This behaviour could be handled using Annotations in the Models.Recipient Class. 
Example
`[Required]
public string Location { get; set; }
`

Using Annotations, the method will return an error if Location is not set.

GiEnJul/Controllers/RecipientController.cs
joffe97: Moved the null value check into the controller.

Sommerli:
I belive some changes will have to be made, with Issue #135. 

I am not sure if this PR should be "on-hold" until 135 is resolved.

Sommerli (approved):

PR #132, opened 2021-06-29
GenericRepository GetAllAsync and GetAllByQueryAsync

### LINKED ISSUE #117: Backend - Add method to fetch all entities

Sommerli:
Closes #117 
Added the possibility to query data from a Table, or get all rows within a table. Also added empty constructors where needed.

Building queries can be done with :
TableQuery.GenerateFilterCondition("PartitionKey", QueryComparisons.Equal, "my_partitionkey")
https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.storage.table.tablequery.generatefiltercondition?view=azure-java-legacy

These queries could be combined using the TableQuery.CombineFilters-method.

Or by simply passing a string "PartitionKey eq 'my_partitionkey'"

I have decided to suppress warnings as 
`TableQuerySegment<T> queryResult = await _table.ExecuteQuerySegmentedAsync(query, token);`  will not compile, without 
`where T : ITableEntity, new()`. 

The compiler will display : 
The type 'typename' must be a non-abstract type with a public parameterless constructor in order to use it as parameter 'parameter' in the generic type or method 'generic'

Without defining `where T : ITableEntity, new()`, queries are unable to convert results to our Entities.classes.

Edit: 
Created the EntityBase class, inheriting from ITableEntity. GenericRepository now expects T type EntityBase. Changed Entities.classes to inherit from EntityBase


elisabethnetland (dismissed):

GiEnJul/Entities/Connection.cs
```diff
@@ -2,19 +2,13 @@
...
+    public class Connection : EntityBase
     {
         public Connection()
-        {
-        }
+        { }
 
-        // RowKey = Guid
-        // PartitionKey = loc_{receiverLocation}
         public Connection(string receiverLocation, string rowKey) : base(receiverLocation, rowKey)
```
Sommerli, line 10: This constructor will be changed when Events are implemented properly.

RonHaland (approved):

Issue #131, opened 2021-06-29
Frontend - Integrate material ui components with the InputValidator component


joffe97:

TheseusDeus:
was closed with #139

PR #139, opened 2021-06-30
Changed how InputValidation is implemented. Added material ui to Inpu…

joffe97:
…tValidation component. Added InputValidation to places it wasn't used.

Fix issue #131

* Changed how InputValidation is implemented. 
* Added material ui to InputValidation component. 
* Added InputValidation to places where validation were implemented without the InputValidation component.


GiEnJul/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
TheseusDeus, line 1: This file should no be in GiEnJul/*

GiEnJul/ClientApp/src/routes/registerGiver/ContactInfo.tsx
```diff
@@ -1,8 +1,10 @@
 import * as React from 'react';
 import { Button, Grid, Container } from '@material-ui/core';
 import useStyles from './Styles';
-import validator from 'validator';
```
TheseusDeus, line 4: Have you checked that all of these changes wont interfere with the signup process and that it still connects to the API?

TheseusDeus (changes requested):
At least remove the package-lock.json file

.gitignore
```diff
@@ -348,3 +348,8 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
+
+# Custom
+package.json
```
TheseusDeus, line 354: cant ignore these files in git. Need them in the /ClientApp/** folder for when implementing libraries

TheseusDeus (changes requested):

.gitignore
joffe97: Fixed this by adding `/` in front of the filename in .gitignore. This way it only ignores package.json and package-lock.json in the root folder. https://stackoverflow.com/questions/3637660/how-to-exclude-file-only-from-root-folder-in-git

TheseusDeus: Thats a great solution

TheseusDeus (approved):

PR #142, opened 2021-07-01
Added Events, and refactored Giver to support events

### LINKED ISSUE #114: Backend - Add event-based entity structure
Entities should be split by their related event, e.g. Jul21 ...

Update entities to support this feature, possibly add events as a table.

### LINKED ISSUE #135: Backend - Refactor Models.Giver to utilize DTO
The current implementation of the models-directory is not correct, as models should contain DTOs related to API-actions.
Models.Giver should be renamed to AddGiverDto, and only contain the necessary properties to submit a new Giver. 
This also applies to Models.Reicipient

Sommerli:
Note: An active event on a location is now required to add a Giver

Added the Event entity and repository.

Refactored controller to utilize PostGiverDto instead of Giver. Also changed tests to reflect new implementation.

Closes #114 
Closes #135

Sorry for the long PR, but implementing Events and refactoring Models.Giver separately would require some unnecessary test-problems.


GiEnJul/Entities/Giver.cs
```diff
@@ -10,21 +10,15 @@ public class Giver : EntityBase
...
-            RowKey = rowKey;
-        }
-
-        public Giver(string location)
-        {
-            RowKey = Guid.NewGuid().ToString();
-            PartitionKey = $"loc_{location}";
-            Location = location;
+            Location = location ?? throw new ArgumentNullException(nameof(location));
+            EventName = eventName ?? throw new ArgumentNullException(nameof(eventName));
```
RonHaland, line 16: Big fan 😎👍

RonHaland (commented):
We could go through this one together.

GiEnJul/Entities/Giver.cs
```diff
@@ -10,21 +10,15 @@ public class Giver : EntityBase
         public Giver()
         { }
 
-        public Giver(string location, string rowKey)
+        public Giver(string location, string eventName) : base($"{eventName}_{location}", Guid.NewGuid().ToString())
```
joffe97, line 13: Could it be an idea to have fks `eloc_{eventName}_{locaction}` as primary key? The `eloc` part would be nice to have if there some time in the future would be added another type of partition key (fks `loc_{location}`).

elisabethnetland (dismissed):

GiEnJul/Entities/Giver.cs
Sommerli: I think this would add more complexity than needed. Givers could be fetched with EventName or Location utilizing the properties: 
image

GiEnJul/Infrastructure/AutoMapperProfile.cs
```diff
@@ -8,26 +8,25 @@ public AutoMapperProfile()
...
+                //.ForMember(dest => dest.SubmitterPhoneNumber, opt => opt.MapFrom(src => src.Receiver.ContactPhoneNumber))
+                //.ForMember(dest => dest.Institution, opt => opt.MapFrom(src => src.Receiver.Institution))
+                //.ForMember(dest => dest.ReferenceId, opt => opt.MapFrom(src => src.Receiver.ReferenceId))
+                //.ForMember(dest => dest.ReceiverLocation, opt => opt.MapFrom(src => src.Receiver.Location))
+                //.ForMember(dest => dest.Dinner, opt => opt.MapFrom(src => src.Receiver.Dinner))
+                //.ForMember(dest => dest.Dessert, opt => opt.MapFrom(src => src.Receiver.Dessert))
+                //.ForMember(dest => dest.Note, opt => opt.MapFrom(src => src.Receiver.Note))
+                //.ForMember(dest => dest.Event, opt => opt.MapFrom(src => src.Receiver.Event))
+                //.ForMember(dest => dest.PersonCount, opt => opt.MapFrom(src => src.Receiver.FamilyMembers.Count));
+
```
RonHaland, line 28: For future ref: 
It's good to write a short comment explaining why this code is disabled and not removed, possibly with reference to the issue that caused this.

RonHaland (approved):

PR #155, opened 2021-07-02
Added confirmation to institution form

### LINKED ISSUE #126: Frontend -Institution page confirmation
When POSTing a form the page should display a confirmation after getting a good response

joffe97:
Fix #126

### DIFF
--- a/GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
+++ b/GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
@@ -7,6 +7,7 @@ import LOCATIONS from '../../common/constants/Locations';
 import Gender from '../../common/enums/Gender';
 import InputValidator from '../InputFields/Validators/InputValidator';
 import { isEmail, isNotNull, isPhoneNumber } from '../InputFields/Validators/Validators';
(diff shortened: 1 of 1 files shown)


GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
```diff
@@ -133,13 +158,26 @@ const RegistrationForm = () => {
             ReferenceId:pid,
             FamilyMembers: personsList
         }
-        fetch('https://localhost:5001/api/recipient', {
+        setIsLoading(true);
+        await fetch('https://localhost:5001/api/recipient', {
```
TheseusDeus, line 162: This should be:  await fetch('api/recipient') instead

TheseusDeus (changes requested):
change the fetch call

GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
joffe97: Yeah that might be a good idea! 👍

TheseusDeus (approved):

PR #158, opened 2021-07-05
Admin edit landing page

### LINKED ISSUE #147: Frontend - edit landingPage
Editor in adminPage where admin can edit the text in the landing page. 
Section for: 
- [x] Hvordan fungerer
- [x] Ofte stilte spørsmål
- [x] Kontakt
- [x] For bedrifter

elisabethnetland:
Issue #147 
Implemented interface for editing the landigpage. The text that is written in the edior is translated to HTML. 
To be able to close #147 tasks to store the text and fetch it to the landingpage has to be done.

### DIFF
--- a/GiEnJul/ClientApp/src/components/admin/editLandingPage/EditCompany.tsx
+++ b/GiEnJul/ClientApp/src/components/admin/editLandingPage/EditCompany.tsx
@@ -7,7 +7,6 @@ import draftToHtml from 'draftjs-to-html';
 import { Typography, FormGroup, TextField } from '@material-ui/core';
 
 type Props = {
-    step: Number,
     companyState: EditorState
     setCompanyState: (input: EditorState) => void,
     companyTitleState: string,
@@ -15,7 +14,7 @@ type Props = {
 }
 
 
-const EditCompany: React.FC<Props> = ({ step, companyState, setCompanyState, companyTitleState, setCompanyTitleState }) => {
+const EditCompany: React.FC<Props> = ({companyState, setCompanyState, companyTitleState, setCompanyTitleState }) => {
     const [content, setContent] = useState<string>('');
 
         return (
--- a/GiEnJul/ClientApp/src/components/admin/editLandingPage/EditCompany.tsx
+++ b/GiEnJul/ClientApp/src/components/admin/editLandingPage/EditCompany.tsx
@@ -18,7 +18,6 @@ type Props = {
 const EditCompany: React.FC<Props> = ({ step, companyState, setCompanyState, companyTitleState, setCompanyTitleState }) => {
     const [content, setContent] = useState<string>('');
 
-    if (step === 3) {
         return (
             <FormGroup>
                 <Typography variant='h6'>Tittel:</Typography>
@@ -53,11 +52,5 @@ const EditCompany: React.FC<Props> = ({ step, companyState, setCompanyState, com
                 />
             </FormGroup>
         )
-    }
-    else {
-        return (
-            <div></div>
-        )
-    }
 }
 export default EditCompany;
--- a/GiEnJul/ClientApp/src/components/admin/editLandingPage/EditCompany.tsx
+++ b/GiEnJul/ClientApp/src/components/admin/editLandingPage/EditCompany.tsx
@@ -8,35 +8,35 @@ import { Typography, FormGroup, TextField } from '@material-ui/core';
 
 type Props = {
     step: Number,
-    howState: EditorState
-    setHowState: (input: EditorState) => void,
-    howTitleState: string,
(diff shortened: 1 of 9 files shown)


GiEnJul/ClientApp/src/components/admin/editLandingPage/EditMacro.tsx
```diff
@@ -1,8 +1,102 @@
...
+                    }}>Kontakt</Button>
+                </Grid>
+                <Grid item xs={2}>
+                    <Button variant='outlined' fullWidth={true} onClick={() => {
+                        setStep(5)
+                    }}>Start "Gi en jul"</Button>
+                </Grid>
+            </Grid>
+                    
+                    <EditHow
```
joffe97, line 68: I think it would be nice to have a switch case statement here that checks which component should render. I think it would be a bit more readable this way, instead of having the if statements inside the components. The components would also be easier to reuse.

GiEnJul/ClientApp/src/components/admin/editLandingPage/EditMacro.tsx
```diff
@@ -1,8 +1,102 @@
...
+    
+    return (
+        <Container>
+            <Typography align="center" variant='h4' >Rediger gienjul.no</Typography>
+            <Grid
+                container
+                direction="row"
+                justify="center"
+                alignItems="center">
+                <Grid item xs={2}>
```
joffe97, line 41: These nav buttons could've been placed inside a for loop, where the button names is looped through. The step variable could have been set by using the loop index.

joffe97 (changes requested):

GiEnJul/ClientApp/src/components/admin/editLandingPage/EditCompany.tsx
```diff
@@ -0,0 +1,63 @@
...
+                    dangerouslySetInnerHTML={{
+                        __html: content
+                    }}
+                />
+            </FormGroup>
+        )
+    }
+    else {
+        return (
+            <div></div>
```
Sommerli, line 59: Burde det vises en eventuell errormsg her?

elisabethnetland: It is removed with the new changes :)

GiEnJul/ClientApp/src/components/admin/editLandingPage/EditMacro.tsx
elisabethnetland: changes implemented

GiEnJul/ClientApp/src/components/admin/editLandingPage/EditCompany.tsx
```diff
@@ -0,0 +1,56 @@
...
+type Props = {
+    step: Number,
+    companyState: EditorState
+    setCompanyState: (input: EditorState) => void,
+    companyTitleState: string,
+    setCompanyTitleState: (input: string) => void,
+}
+
+
+const EditCompany: React.FC<Props> = ({ step, companyState, setCompanyState, companyTitleState, setCompanyTitleState }) => {
```
joffe97, line 18: The step variable can probably be removed now that it's not used in the component anymore. This applies to all these "subpage" components.

GiEnJul/ClientApp/src/components/admin/editLandingPage/EditMacro.tsx
```diff
@@ -29,74 +30,67 @@ export default () => {
...
-                    contactTitleState = {contactTitleState}
-                    setContactTitleState ={setContactTitleState}/>
-                    <EditStart
-                    step = {step}
-                    startState ={startState}
-                    setStartState ={setStartState}
-                    startTitleState = {startTitleState}
-                    setStartTitleState ={setStartTitleState}/>
-
+            {(step === 0 &&
```
joffe97, line 56: I think this should start with 1 instead of 0. This applies to the next condition here as well (step === 1)

joffe97 (changes requested):

GiEnJul/ClientApp/src/components/admin/editLandingPage/EditMacro.tsx
elisabethnetland: the setStep function is set by the index of the sections array, and therefore my opinion is that it is natural to start at 0. If it starts at 1 the setStep would be setStep(index+1).

joffe97: Okay, I agree that this is more natural 👍

joffe97 (approved):
Nice work!

PR #159, opened 2021-07-05
Added material ui to locations, persons, dinner and dessert forms in …

joffe97:
…instituion registration page


Sommerli:
Should package-lock be included?

Sommerli (approved):
Maybe package-lock should not be included? otherwise all good

joffe97:
> Maybe package-lock should not be included? otherwise all good

The package-lock file updated itself when I downloaded a necessary package. It also got updated to package-lock v2, which is backwards compatible with package-lock v1 (which it was before). This should therefore not affect master in a negative way.

Sommerli (approved):

PR #167, opened 2021-07-06
Added select component that uses native select design on mobile

### LINKED ISSUE #162: Frontend - Add select component
Add select component that uses native select on mobile, and material ui for desktop.

joffe97:
Fix #162

### DIFF
--- a/GiEnJul/ClientApp/src/common/functions/IsMobile.tsx
+++ b/GiEnJul/ClientApp/src/common/functions/IsMobile.tsx
@@ -1,0 +1,3 @@
+export const isMobile = () => {
+    return window.innerWidth < 576;
+}
--- a/GiEnJul/ClientApp/src/common/functions/IsMobile.tsx
+++ b/GiEnJul/ClientApp/src/common/functions/IsMobile.tsx
@@ -1,3 +1,22 @@
 export const isMobile = () => {
-    return window.innerWidth < 576;
-}
+	let hasTouchScreen = false;
+	if ("maxTouchPoints" in navigator) {
+		hasTouchScreen = navigator.maxTouchPoints > 0;
+	} else if ("msMaxTouchPoints" in navigator) {
+		hasTouchScreen = navigator.msMaxTouchPoints > 0;
+	} else {
+		let mQ = matchMedia("(pointer:coarse)");
+		if (mQ && mQ.media === "(pointer:coarse)") {
+			hasTouchScreen = !!mQ.matches;
+		} else if ("orientation" in window) {
+			hasTouchScreen = true; // deprecated, but good fallback
+		} else {
+			// Only as a last resort, fall back to user agent sniffing
+			let UA = navigator.userAgent;
+			hasTouchScreen =
+				/\b(BlackBerry|webOS|iPhone|IEMobile)\b/i.test(UA) ||
+				/\b(Android|Windows Phone|iPad|iPod)\b/i.test(UA);
+		}
+	}
+	return hasTouchScreen;
+};
--- a/GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
+++ b/GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
@@ -1,10 +1,4 @@
-﻿import {
-	BaseTextFieldProps,
-	FormControlTypeMap,
-	InputBaseProps,
-	TextField,
-} from "@material-ui/core";
-import { SelectInputProps } from "@material-ui/core/Select/SelectInput";
+﻿import { TextField } from "@material-ui/core";
 import * as React from "react";
 import { FC, useEffect, useState } from "react";
 import SelectInput from "../SelectInput";
@@ -29,7 +23,7 @@ interface InputValidatorProps {
 	viewErrorTrigger?: number;
 
 	value: any;
-	onChange: any;
+	onChange: (e: any) => void;
 	label: string;
 	name: string;
 	type?: string;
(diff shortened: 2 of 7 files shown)


GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
```diff
@@ -38,10 +38,10 @@ interface InputValidatorProps {
     autoFocus?: boolean,
     size?: any,
     multiline?: boolean,
-    rowsMax?: number | string
+    rowsMax?: number | string,
+    placeholder?: string,
 
     options?: {value: any, text: string}[],  // Only for "type=select".
-    isMobile?: boolean,   // Only for "type=select". If true: select dropdown uses native dropdown, which is better for mobile.
```
elisabethnetland, line 44: I think textFields also should have a layout that is optimized for mobile. For example that the placeholderText is removed if the placeholderText exceeds the input field. 
This could be another issue :)
image

joffe97: Hmm good point. The text should probably also be shortened in this case. Could it be an idea to wait with this until the design for it is finished? Khoi might have a nice way to deal with this problem.

elisabethnetland: Sounds good, I will create an issue about this.

TheseusDeus: Is it supposed to be optimized for phone at all. Is this site not just for admins that register families?

joffe97: That's true. This component (InputValidator) is also used in the giver form though. A nice solution could be to truncate the placeholders, and/or set overflow to none.

GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
```diff
@@ -85,7 +86,8 @@ const InputValidator: FC<InputValidatorProps> = (
     }
 
     const validate = () => {
-        let errorMsg = "";
+        let errorMsg: string | undefined = undefined;
+        let isAnyError = false;
         for (let i = 0; i < validators.length; i++) {
```
Antonio-Cabrero, line 91: if validators is an array of values its better to use 
validators.map(validator => { // do something })

for better readability

GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
```diff
@@ -85,7 +86,8 @@ const InputValidator: FC<InputValidatorProps> = (
...
             let isValid = validators[i](value);
             switch (typeof setIsValids) {
```
Antonio-Cabrero, line 93: clean up?

GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
```diff
@@ -119,41 +123,32 @@ const InputValidator: FC<InputValidatorProps> = (
 
     switch (type) {
         case "select":
+            if (!options) throw Error("Options must be defined in select input");
```
Antonio-Cabrero, line 126: Error handling should be managed in the function which is responsible

GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
```diff
@@ -119,41 +123,32 @@ const InputValidator: FC<InputValidatorProps> = (
...
+                    error={isError}
+                    autoFocus={autoFocus}
+                />
             )
         default:
             return (
                 <TextField
-                    error={!!error}
-                    helperText={errorMessages === undefined ? undefined : error}
+                    error={isError}
```
Antonio-Cabrero, line 150: hasError makes more sense :P

GiEnJul/ClientApp/src/common/functions/IsMobile.tsx
```diff
@@ -0,0 +1,3 @@
+export const isMobile = () => {
+    return window.innerWidth < 576;
```
Antonio-Cabrero, line 2: this maybe cause bugs as the users innerWidth can exceed 576 in landscape mode for many devices.
a more thorough check is necessary. Look into MDN for ways to check if a device is mobile device :)

GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
```diff
@@ -97,10 +99,12 @@ const InputValidator: FC<InputValidatorProps> = (
                     break;
             } 
             if (!isValid && viewError) {
-                errorMsg = errorMessages !== undefined ? errorMessages[i] : "dummy";
+                errorMsg = errorMessages !== undefined ? errorMessages[i] : undefined;
+                isAnyError = true;
             }
         }
-        setError(errorMsg);
+        setIsError(isAnyError);
```
Antonio-Cabrero, line 106: double setState calls can cause rerender problems, try to avoid it. 
a good way to do this is to create a state that has everything to do with "validating" or "errorState", 
you create an object with the different properties and then setErrorState as needed like so:

if (isValid) {
   setErrorState((prev) => ({ ...prev, hasError: false })) 
} else if (!isValid && errorMessage !== undefined) {
   setErrorState((prev) => ({ 
                    ...prev, hasError: false,
                               errorMessage: errorMsg,
 }))

Antonio-Cabrero (commented):
Great job, just minor fixes for best practices and shinnier code :)

joffe97:
> Great job, just minor fixes for best practices and shinnier code :)

I've changed the code based on the suggestions from @Antonio-Cabrero. 
There could still be some improvements:
* Change all occurrences of `any` into a suitable type.
* Gather states into object states.
These is rather big tasks and should be done in separate issues.

TheseusDeus (approved):
Looks really good, esecially the new isMobile function

PR #178, opened 2021-07-06
Change return types for RecipientController and RecipientRepository

Sommerli:
RecipientController.PostAsync now returns [OkResult](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.ok?view=aspnetcore-5.0#Microsoft_AspNetCore_Mvc_ControllerBase_Ok) instead of CreatedAtAction, as the returned object is never used. 

RecipientRepository now return model-objects instead of entities.


GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
```diff
@@ -177,7 +177,7 @@ const RegistrationForm = () => {
             body: JSON.stringify(submit)
         })
         .then((response) => {
-            if (response.status === 201) {
+            if (response.status === 200) {
```
Sommerli, line 180: Could be changed to 
```suggestion
            if (response.ok) {
```
?

TheseusDeus (approved):
looks good

GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
TheseusDeus: dont really need it, but is better for readability

PR #179, opened 2021-07-06
Added alert component. Added alert component to institution form

### LINKED ISSUE #169: Frontend - Create alert component
Add an alert component using material ui.

joffe97:
Fix #169 
Decided to go with the solution proposed by @Sommerli instead, and use the native material ui snackbar component. The refactoring of the states wasn't done in this issue, as this is a rather big issue.


GiEnJul/ClientApp/src/components/alert/AlertHover.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const [viewAlert, setViewAlert] = useState(false);
+    const [,setTimer] = useState<NodeJS.Timeout | undefined>();
+
+    const extendedOnClose = (event: any) => {
+        setViewAlert(false);
+        if (onClose) onClose(event);
+    }
+
+    useEffect(() => {
+        if (!children) return;
```
Antonio-Cabrero, line 43: a best practice is to try to avoid "negative" or "falsy" statements.

GiEnJul/ClientApp/src/components/alert/AlertHover.tsx
```diff
@@ -0,0 +1,71 @@
...
+        setViewAlert(true);
+        setTimer(t => {
```
Antonio-Cabrero, line 45: t is not descriptive, always try to write readable names.

GiEnJul/ClientApp/src/components/alert/AlertHover.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    const [viewAlert, setViewAlert] = useState(false);
+    const [,setTimer] = useState<NodeJS.Timeout | undefined>();
+
+    const extendedOnClose = (event: any) => {
+        setViewAlert(false);
+        if (onClose) onClose(event);
+    }
+
+    useEffect(() => {
```
Antonio-Cabrero, line 42: I think this useEffect needs clean up. Logic is a bit "vague", and what I mean by that is that is code seems to not be focused on whos gonna read it later. Always try to make your code maintainable.

GiEnJul/ClientApp/src/components/alert/AlertHover.tsx
```diff
@@ -0,0 +1,71 @@
...
+        if (children && !setChildren) {
+            throw Error("Cannot use the children property without the setChildren property. See documentation.");
+        }
+    })
+
+    return (
+        <Alert 
+        id={`_alertHover ${id ? id : ""}`}
```
Antonio-Cabrero, line 62: this could be made into a const which checks the truthyness instead of inline, to improve readability

GiEnJul/ClientApp/src/components/alert/AlertHover.tsx
```diff
@@ -0,0 +1,71 @@
...
+        className={`${classes.root} ${className} ${!viewAlert ? classes.hide : ""}`}
```
Antonio-Cabrero, line 63: same as above, or implement classnames library

GiEnJul/ClientApp/src/components/alert/AlertHover.tsx
```diff
@@ -0,0 +1,71 @@
...
+const AlertHover: FC<IAlertHover> = ({
+    fadeTime,
+    className,
+    children,
+    setChildren,
+    id,
+    onClose,
+    ...rest
+}) => {
+    const classes = useStyles();
```
Antonio-Cabrero, line 32: useStyels entales this is a hook, however i only see classes specifically for this component. ( i may be wrong) However it doesnt seem necessary to write it separately unless youll be exporting different classses for different components from it. We can go over this.

Antonio-Cabrero (commented):
Good work here,
Id jsut advise to move away from using more that 3 useStates. In the file InstitutionForm.tsx, you can group things into a big state obj. The resetForm() will experience bugs at some point because there is way too many setState calls, usually you dont want more than 2 setState calls within the same closure function.

I wrote how to make a bigger state obj on the other pr. #167

Sommerli (commented):
Look into the possibility to resolve this functionality with snackbars. https://material-ui.com/components/snackbars/

GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
```diff
@@ -1,364 +1,433 @@
...
+	const [isValidContactPhoneNumber, setIsValidContactPhoneNumber] =
+		useState(false);
+
+	const [contactEmail, setContactEmail] = useState("");
+	const [isValidContactEmail, setIsValidContactEmail] = useState(false);
+
+	const [isLoading, setIsLoading] = useState(false);
+	const [alertMsg, setAlertMsg] = useState<React.ReactNode>("");
+	const [alertSeverity, setAlertSeverity] = useState<any>();
+	const [alertOpen, setAlertOpen] = useState(false);
```
Sommerli, line 76: I think these properties could be combined to a type, containing all variables needed to fill the snackbar.

Sommerli (changes requested):
Let me know what you think about the comment :)

GiEnJul/ClientApp/src/components/institution/InstitutionForm.tsx
joffe97: I agree! But I think this should be done along with the refactoring all the states in this file.

Sommerli (approved):

PR #186, opened 2021-07-07
Roleretrival

### LINKED ISSUE #184: Frontend -Role retrieval
Fetch token from managment API for Auth0
Use this to get the user object and display different things on the adminpage based on role

TheseusDeus:
When logging into the admin page a user will see content based on roles
Fetches a token from the auht0 managment API after the user is logged in. This token is then used as authorization when asking for the user object from the mAPI. 
Data is then stored in state
Fixed warnings that appeared in the console for attributes
Closes #184

### DIFF
--- a/GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
+++ b/GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
@@ -1,23 +1,80 @@
-import { withAuthenticationRequired } from "@auth0/auth0-react";
+import { useAuth0, withAuthenticationRequired } from "@auth0/auth0-react";
 import { Container } from "@material-ui/core";
 import * as React from "react";
+import { useEffect, useState } from "react";
 import LoadingPage from "../../common/components/LoadingPage";
+import InstitutionMacro from "../institution/InstitutionMacro";
 import LogOutButton from "../login/LogOutButton";
 import AdminMenu from "./common/AdminMenu";
 import Completed from "./connections/Completed";
 import Suggested from "./connections/Suggested";
 
 function AdminPage() {
-  return (
-    <>
-      <LogOutButton></LogOutButton>
-      <Container maxWidth="xl">
-        <AdminMenu />
-        <Suggested />
-        <Completed />
-      </Container>
-    </>
-  );
+  const { user, getAccessTokenSilently } = useAuth0();
+  const [appMetadata, setappMetadata] = useState(null);
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  useEffect(() => {
+    const getappMetadata = async () => {
+      const domain = "dev-r7fmessb.eu.auth0.com";
+
+      try {
+        console.log("accessing token");
+        const accessToken = await getAccessTokenSilently({
+          audience: `https://${domain}/api/v2/`,
+          scope: "read:current_user",
+        });
+        console.log(accessToken);
+
+        const userDetailsByIdUrl = `https://${domain}/api/v2/users/${
+          user!.sub
+        }`;
+        console.log("url: " + userDetailsByIdUrl);
+
+        const metadataResponse = await fetch(userDetailsByIdUrl, {
+          headers: {
+            Authorization: `Bearer ${accessToken}`,
+          },
+        });
+        console.log("response: ");
+        console.log(metadataResponse);
+
+        const user_appMetadata = await metadataResponse.json();
+        console.log(user_appMetadata.app_metadata);
+        const userMETA = user_appMetadata.app_metadata;
+        setappMetadata(userMETA);
+        setIsLoaded(true);
+      } catch (e) {
+        console.log(e.message);
+        setIsLoaded(true);
+      }
+    };
+
+    getappMetadata();
+  }, [user,getAccessTokenSilently]);
+
+  if (!isLoaded) {
+    return <LoadingPage />;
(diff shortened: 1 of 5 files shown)


GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
```diff
@@ -1,23 +1,80 @@
...
+    return <LoadingPage />;
+  }else if(isLoaded && appMetadata == null){
+    return <LoadingPage/>
+}else {
+    if (appMetadata!["role"] === "institution") {
+      return <InstitutionMacro />;
+    } else if (appMetadata!["role"] === "admin") {
+      return (
+        <>
+          <LogOutButton></LogOutButton>
```
Antonio-Cabrero, line 60: ```<LogOutButton />``` no need to open it :)

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
```diff
@@ -1,23 +1,80 @@
...
-      </Container>
-    </>
-  );
+  const { user, getAccessTokenSilently } = useAuth0();
+  const [appMetadata, setappMetadata] = useState(null);
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  useEffect(() => {
+    const getappMetadata = async () => {
+      const domain = "dev-r7fmessb.eu.auth0.com";
```
Antonio-Cabrero, line 19: this could be stored somewhere to be globally available

Antonio-Cabrero (commented):
Good work, minor changes

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
TheseusDeus: yea, I already have, hehe. Just forgot to use it here.
Have it stored in a .env file

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
TheseusDeus: trueee

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
```diff
@@ -1,23 +1,80 @@
...
-        <Completed />
-      </Container>
-    </>
-  );
+  const { user, getAccessTokenSilently } = useAuth0();
+  const [appMetadata, setappMetadata] = useState(null);
+  const [isLoaded, setIsLoaded] = useState(false);
+
+  useEffect(() => {
+    const getappMetadata = async () => {
```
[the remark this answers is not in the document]
TheseusDeus: Looks like most people use the axios library for fetching from APIs, so it is probs best to implement this here and all other places where a "fetch()" has been used

Tried doing it like that with the useEffect(myfunc, [depend]), but could not quite get it to work.  It is probs the better solution tho

elisabethnetland (dismissed):

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
```diff
@@ -1,23 +1,89 @@
...
+  useEffect(() => {
+      getUserInformation().then((response: UserDataType) => setUserData(response)),
+      setIsLoaded(true)
+  },[]);
+
+
+
+  if (!isLoaded) {
+    return <LoadingPage />;
+  } else if (isLoaded && userData == null) {
```
Sommerli, line 67: Could be placed with initial "if" as they return same value

TheseusDeus: Not really. That is if you are Loaded and the code couldn't fetch the userData. So it should display an errorpage, but we don't have an errorpage yet. So the return is just a temp

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
```diff
@@ -1,23 +1,89 @@
...
 import * as React from "react";
+import { useEffect, useState } from "react";
 import LoadingPage from "../../common/components/LoadingPage";
+import InstitutionMacro from "../institution/InstitutionMacro";
 import LogOutButton from "../login/LogOutButton";
 import AdminMenu from "./common/AdminMenu";
 import Completed from "./connections/Completed";
 import Suggested from "./connections/Suggested";
 
+type UserDataType = {
```
Sommerli, line 12: Should all variables be nullable? Or should we require certain variables from the 
`getUserInformation().then((response: UserDataType) => setUserData(response)),` to actually be set?

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
Sommerli: I see 👍 then its probably a good idea to throw an exception or handle the error in a temporary way, instead of leaving a "functional" endpoint where it should fail.

TheseusDeus: Fixed with d931b9c

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
TheseusDeus: Depends on what variables we really need

Sommerli (approved):
> Depends on what variables we really need

This should be taken into consideration, but if its not addressed in this issue, the PR can be completed

PR #189, opened 2021-07-08
Replaced static LOCATIONS with function GetLocations

Sommerli:
Forms that require a location now fetch active locations from Table Storage.
Example
image


GiEnJul/ClientApp/src/components/registerGiver/LocationGiver.tsx
```diff
@@ -5,95 +5,95 @@ import { ValidatorForm, SelectValidator } from 'react-material-ui-form-validator
...
-          </Grid>
-          <Grid item>
-            <Button variant="contained" type="submit">Neste</Button>
-          </Grid>
-        </Grid>
-      </ValidatorForm>
-    </Container>
-  )
 }
 export default LocationGiver
```
Sommerli, line 99: This file is unchanged.

GiEnJul/ClientApp/src/components/institution/InstitutionLocations.tsx
```diff
@@ -36,12 +36,17 @@ const Location:FC<LocationProps> = ({
         setIsErr(viewError && !isValid);
     }, [value, setIsValidLocation, setIsErr, viewError])
 
-    return(
+    const [locationList, setLocationList] = useState<string[]>(['Ingen lokasjoner']);
```
joffe97, line 39: I think it would be nice to initialize this with an empty array, instead of "Ingen lokasjoner". This way the "Ingen lokasjoner" wouldn't be used as a location if no active events. I would suggest to change the options in the SelectValidator for cases with no locations, and use an empty string as value. This way the user can't proceed if "Ingen lokasjoner" is chosen.

Edit: I see that the "Ingen lokasjoner" will never be used, because it's overwritten by the useEffect below. The handling of empty arrays should be done though.

Sommerli: Giver: Currently a spinner is shown for the dropdown as the list is waiting to be populated.

Institution: A text displaying that no locations are currently active is shown, and that the user should try again later.

GiEnJul/ClientApp/src/common/constants/Locations.tsx
```diff
@@ -1,9 +1,6 @@
-const LOCATIONS: string[] = [
-    'Stavanger',
-    'Sandnes',
-    'Bodø',
-    'Sola',
-    'Nittedal'
-].sort();
+async function getLocations() {
+    const locations = await fetch('/api/event/ActiveLocations').then(response => response.json()).then(data => { return data });
```
TheseusDeus, line 2: Formatting: Put the then()s on a newline each for readability

TheseusDeus (approved):
Good job. Fix the merge conflict and you'll be good

GiEnJul/ClientApp/src/components/registerGiver/SummaryRegistration.tsx
```diff
@@ -122,8 +121,8 @@ const SummaryRegistration: React.FC<Props> = ({ nextStep, prevStep, handleLocati
                             errorMessages={['Vennligst velg lokasjon']}
                             validators={[isNotNull]}
                             setIsValids={setIsValidLocation}
-                            options={LOCATIONS.map(x => 
-                                {return {value: x, text: x};})}
+                            options={locationOptions.map(x => { return { value: x, text: x }; })}
+                            isMobile={false}
```
Sommerli, line 125:
```suggestion
```

TheseusDeus (approved):

PR #197, opened 2021-07-09
formatted a lot of files with prettier

TheseusDeus:
Ran prettier on a lot of our old files
Tried removing redux, but the router "connectedrouter" takes use of redux so that proved a challenge
May need to change the router to something else then


Sommerli:
Please provide how prettier is implemented, so we all could start using it :)

Sommerli (approved):

TheseusDeus:
> Please provide how prettier is implemented, so we all could start using it :)

image
First download the "Prettier" extension in vs code. Then in ur usersettings make a tab for "[typescriptreact]" and add a defaultformatter. To apply formatting now, just press CTRL+SHIFT+P and search format document or add another option so it will automatically format on save. Line:14 @Sommerli @elisabethnetland

PR #199, opened 2021-07-09
Changed the access modifiers of GenericRepository, and removed the GenericRepo interface

Sommerli:
Added protected to all GenericRepository-functions. Removed IGenericRepository as its implementation is not needed, and was preventing changes to access modifiers.

This should provide better maintainability to our usage of repositories.

Lets say you are using the GiverRepository in a controller, you now only have access to IGiverRepository's functions.

image

### DIFF
--- a/GiEnJul/Controllers/AdminController.cs
+++ b/GiEnJul/Controllers/AdminController.cs
@@ -37,8 +37,8 @@ namespace GiEnJul.Controllers
         // }
         
         [HttpGet("givers")]
-        public async Task<List<Models.Giver>> GetGiversAsync() {
-            return _mapper.Map<List<Models.Giver>>(await _giverRepository.GetAllAsync());
+        public async Task<List<Giver>> GetGiversAsync() {
+            return _mapper.Map<List<Giver>>(await _giverRepository.GetAllAsModelAsync());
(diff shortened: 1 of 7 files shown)


joffe97 (dismissed):
Nice!

RonHaland (approved):

GiEnJul/Controllers/AdminController.cs
```diff
@@ -37,8 +37,8 @@ public AdminController(IEventRepository eventRepository, IGiverRepository giverR
         // }
         
         [HttpGet("givers")]
-        public async Task<List<Models.Giver>> GetGiversAsync() {
-            return _mapper.Map<List<Models.Giver>>(await _giverRepository.GetAllAsync());
+        public async Task<List<Giver>> GetGiversAsync() {
+            return _mapper.Map<List<Giver>>(await _giverRepository.GetAllAsModelAsync());
```
joffe97, line 41: Is the mapping to List<Giver> necessary? If I understand correctly, the GetGiversAsync() function could return IEnumerable instead of List, and the mapping could be removed.

Sommerli: You got it right! Need to fix that

joffe97 (approved):

PR #200, opened 2021-07-09
Cleaned up giver registration. Collected all giver states into bigger…

joffe97:
… object states.


GiEnJul/ClientApp/src/components/registerGiver/IGiverInputs.tsx
```diff
@@ -0,0 +1,9 @@
+interface IGiverInputs {
```
Sommerli, line 1: In my opinion the name of this object does not clearly represent its use. 
Maybe it could be renamed to contain Post/Form/Data.. or something like that.
IGiverFormData might be an option.

Also, should these values be nullable?

GiEnJul/ClientApp/src/components/registerGiver/LocationGiver.tsx
```diff
@@ -1,99 +1,137 @@
...
+interface Props {
+  nextStep: () => void;
+  prevStep?: () => void;
+  values: IGiverInputs;
+  handleLocationChange: (event: any) => void;
+  placeHolder: string;
 }
-const LocationGiver: React.FC<Props> = ({ nextStep, prevStep, handleLocationChange, values, options, placeHolder }) => {
 
+const LocationGiver: React.FC<Props> = ({
```
Sommerli, line 21: LocationGiver should be renamed, as this component renders both Familiy-selection and Location-selection

Sommerli (commented):
Left some comments for an initial review, will continue at a later time 😄

GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
```diff
@@ -1,193 +1,194 @@
 ﻿import { TextField } from "@material-ui/core";
 import * as React from "react";
+import { useCallback } from "react";
 import { FC, useEffect, useState } from "react";
 import SelectInput from "../SelectInput";
 
 interface InputValidatorProps {
```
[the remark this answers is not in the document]
joffe97: I agree that the InputValidator adds some complexity to our solution, but I think the pros of using it overcome the cons.
Some pros is listed below:
- Custom error messages for each validation check. E.g. one for invalid email and another for no email.
- The isValid variables. These make it possible to make more flexible validation and feedback. In the institution form these is for e.g. the dinner->annet field.
- The material ui TextField doesn't support pattern. Also the "tel" type doesn't support Norwegian telephone numbers (in the way we write them).

There is a way to use material ui with validation though by using the react-material-ui-form-validator package. The problem with this package is that you can't add custom validations. There is also no support for solutions like the one in dinner-annet on the institution page. Because of these facts I think the InputValidator is more suitable.

GiEnJul/ClientApp/src/common/constants/FamilySizes.tsx
```diff
@@ -0,0 +1,5 @@
+export const FAMILY_SIZES: { value: number; text: string }[] = [
+  { value: 2, text: "Liten familie" },
+  { value: 5, text: "Vanlig familie" },
```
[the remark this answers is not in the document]
joffe97: Agree! I think we should wait to change this until we get the design though, as we probably have to change it then anyway.

GiEnJul/ClientApp/src/components/registerGiver/IGiverInputs.tsx
joffe97: Agree! Also these should not be nullable 👍

GiEnJul/ClientApp/src/components/registerGiver/LocationGiver.tsx
joffe97: I agree. I think these two forms should be two different components.

GiEnJul/ClientApp/src/components/registerGiver/ContactInfo.tsx
```diff
@@ -1,103 +1,153 @@
...
+    for (let isValid in isValidsState) {
+      if (isValidsState[isValid]) continue;
+      return setState((prev) => {
+        return { ...prev, viewErrorTrigger: prev.viewErrorTrigger + 1 };
+      });
     }
+    nextStep(e);
+  };
+
+  const getValiditySetter = (target: string) => {
```
Antonio-Cabrero, line 59: While its nice to send confirmation that something is or isnt valid, for better flow in the form usually we want to make sure that validation happens when we need something from the user before moving to the next step. This function to me sounds like at the end before the user submits its form, thats when you'll let them know that something is wrong. Imagine that scenario. Just something to think about.

GiEnJul/ClientApp/src/components/registerGiver/LocationGiver.tsx
```diff
@@ -1,100 +1,89 @@
...
-                    label="Lokasjon*"
-                    name="location-input"
-                    value={values.location ? values.location : ""}
-                    id="location-input"
-                    onChange={handleLocationChange}
-                    errorMessages={['Hvor vil du spre glede?']}
+  return (
+    <Container>
+      <ValidatorForm
+        onSubmit={nextStep}
```
Antonio-Cabrero, line 32: When a form is to be submitted it should have a onSubmit function. Having a function that can potentially (if theres a bug) lead the user to submit a form without it being complete, can cause problems.

If im not mistaken you guys are "waiting" for the counter to be complete or 5/5 ie. and then you check if everything is valid and then if it is then the information is submitted, correct?

GiEnJul/ClientApp/src/components/registerGiver/SignUp.tsx
```diff
@@ -1,179 +1,188 @@
...
+import Confirmation from "./Confirmation";
+import ContactInfo from "./ContactInfo";
+import LocationGiver from "./LocationGiver";
+import SummaryRegistration from "./SummaryRegistration";
+import { useEffect, useState } from "react";
+import IGiverFormData from "./IGiverFormData";
+import FamilySizeGiver from "./FamilySizeGiver";
+import getLocations from "../../common/constants/Locations";
+
+const initInputsState: IGiverFormData = {
```
Antonio-Cabrero, line 12: Try matching the type to the name of your variables, leads to less confusion :)

GiEnJul/ClientApp/src/components/registerGiver/SignUp.tsx
```diff
@@ -1,179 +1,188 @@
...
+  const [locationOptions, setLocationOptions] = useState<string[]>([]);
+
+  useEffect(() => {
+    getLocations().then((locationArray) => setLocationOptions(locationArray));
+  }, []);
+
+  // go back to previous step
+  const prevStep = (event: React.FormEvent) => {
+    event.preventDefault();
+    setState((prev) => {
```
Antonio-Cabrero, line 38: missing () around the state obj, i see that everywhere the setState call was use has the same issue, be careful. Typescript wont tell you its wrong but it can cause bugs.

GiEnJul/ClientApp/src/components/registerGiver/SignUp.tsx
```diff
@@ -1,179 +1,188 @@
...
+  };
 
-    const values = { location, fullname, email, phoneNumber, maxRecievers, familyType, confirmationOK  };
-    const submit = { location, fullname, email, phoneNumber, maxRecievers, confirmationOK }
-    const familiyOptions = ['Liten familie', 'Vanlig familie', 'Stor familie'];
+  const handleFamilyChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    setInputsState((prev) => {
+      return { ...prev, maxReceivers: event.target.value };
+    });
+  };
```
Antonio-Cabrero, line 81: All handleChange functions can be rewritten in a much simpler way, this is just making things less repetitive.

const handleOnChange = (target, value) => {
   setState((prev) => ({ ...prev, [target]: value, })) 
}

Notice no return function needed.

GiEnJul/ClientApp/src/components/registerGiver/SignUp.tsx
```diff
@@ -1,179 +1,188 @@
...
-                <Container>
+  const handleConfirm = (submitOK: boolean) => {
+    setState((prev) => {
+      return { ...prev, confirmationOK: submitOK };
+    });
+  };
 
-                </Container>
-            )
-    };
```
Antonio-Cabrero, line 177: Theres a lot of unnecessary rewriting of Container tagm CSSBaseline tag and the div with classname paper. Make the switch to return something wrapped in a fragment (<></>) and then just set whatever is returned into the other three elements.

Sommerli (commented):
Got some suggestions here, mainly to shorten down on unused InputValidator-fields.

After our talk some filenames could also be due for changes.

GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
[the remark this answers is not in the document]
Sommerli: Maybe @Antonio-Cabrero got some input on this.

After our talk we do see that InputValidator handles much of the operations that native components could do, but it provides some added flexibility.

Antonio-Cabrero (commented):
Very nice work guys, I'm very impressed. I left some comments to check out. All quick fixes. I do want to stress once again that the naming of variables is very important and must be descriptive. Try avoiding generic naming as well as abbreviations. 
Also in line 62 of the file SelectInput, i loved the conditional rendering you guys used there. Im not sure about this but i think you can (if you want, not necessary) write: 

const args = { 
                key={`n_${name}_${o.text}`}
                value={o.value}
                className="text-capitalize"
}
and just pass it down to each tag like so: <option {...args} />
However, im not completely sure haha :P

All in all great work!

GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
```diff
@@ -1,193 +1,194 @@
...
+  viewErrorTrigger,
+  value,
+  onChange,
+  type,
+  label,
+  name,
+  id,
+  className,
+  disabled,
+  margin,
```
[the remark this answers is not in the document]
joffe97: I checked this out. This is a fact for some components, but the majority (mostly in institution registration) doesn't have margin set.

GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
```diff
@@ -1,193 +1,194 @@
...
+  variant,
+  autoComplete,
+  fullWidth,
+  autoFocus,
+  size,
+  multiline,
+  rowsMax,
```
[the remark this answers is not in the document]
joffe97: Size is used in institution/FormFood.tsx and rowsMax is used in institution/InstitutionForm.tsx.

GiEnJul/ClientApp/src/components/registerGiver/IGiverFormData.tsx
```diff
@@ -0,0 +1,9 @@
+interface IGiverFormData {
+  location: string;
+  fullname: string;
+  email: string;
+  phoneNumber: string;
+  maxReceivers: string;
```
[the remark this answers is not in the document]
joffe97: I checked into this and concluded that it's best to keep it as a string. This is because it's an select tag. When no item is selected the value is set to "". This could maybe be set to undefined, but I think this will make some problems with react (uncontrolled variable). Because of this empty string, I think that the value always should be a string.

GiEnJul/ClientApp/src/components/InputFields/Validators/InputValidator.tsx
```diff
@@ -1,193 +1,194 @@
...
+  type,
+  label,
+  name,
+  id,
+  className,
+  disabled,
+  margin,
+  variant,
+  autoComplete,
+  fullWidth,
```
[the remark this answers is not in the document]
joffe97: This is not true on all fields in institution registration page. I think this should be kept as it is, so it stays the same as the TextField component.

GiEnJul/ClientApp/src/components/registerGiver/Macro.tsx
```diff
@@ -0,0 +1,151 @@
...
+  phoneNumber: "",
+  maxReceivers: "",
+};
+
+const initState: { step: number; confirmationOK?: boolean } = {
+  step: 1,
+  confirmationOK: undefined,
+};
+
+const SignUp = () => {
```
TheseusDeus, line 24: I like the renaming from SignUp to Macro, but should it be more descriptive? SignUpMacro | RegistrationMacro, or something along those lines?
line 24 has to be changed to match the name of the file. Same as the export at line 151

TheseusDeus (changes requested):

GiEnJul/ClientApp/src/components/registerGiver/Macro.tsx
joffe97: Yeah I thought about this, but I feel that Macro is enough, because everything in the folder is for the registration/signup.

joffe97: The component should have another name though, as it's used elsewhere as well. I'll change it to RegistrationMacro.

TheseusDeus: Yea, sounds like a good plan

TheseusDeus: c76abe1 only removed unecessary props. Did you forget a file?

joffe97: Whoops I did 😱 Will fix!

Sommerli (approved):

TheseusDeus (approved):

PR #204, opened 2021-07-12
UseFetch

TheseusDeus:
Added a useFetch() hook in the new hooks folder. This is parallel with the components folder
If fetching something from the backend, import this hook and use the response var in the code


elisabethnetland (approved):
Looks very good, have you tested that it works?

TheseusDeus:
> Looks very good, have you tested that it works?

Yea it should work. I have tested it locally with emulated azure table storage

PR #207, opened 2021-07-13
Usepost

TheseusDeus:
tested it locally for giver and it works
Can now just use this to POST to the DB


Sommerli (approved):

GiEnJul/ClientApp/src/hooks/usePost.tsx
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useState } from "react";
+
+const backend = window.location.origin + "/api";
+
+/**
+ * Returns a response from the backend as a json.
+ * @param url which controller to access and where to POST "/admin/givers" etc.
+ * @param data object or array to POST
+ */
+const usePost = <T extends any>(url: string, data: any) => {
```
joffe97, line 10: Is it necessary to use any here? If any data could be passed in, unknown should be used instead.

PR #208, opened 2021-07-13
Add persons to recipient model

### LINKED ISSUE #205: Backend - Add persons to recipient model
Need a function that gets all persons per recipientID in personRepo.

elisabethnetland:
Closes #205 
Added:
- A query function in person repository creates a list of all persons in the same family by the same recipientID. 
- Mapping from entity to model of persons
- Functionality in admin controller for adding persons to recipients.

### DIFF
--- a/GiEnJul/Controllers/AdminController.cs
+++ b/GiEnJul/Controllers/AdminController.cs
@@ -46,7 +46,8 @@ namespace GiEnJul.Controllers
             var recipients = await _recipientRepository.GetAllAsModelAsync();
             foreach (var family in recipients)
             {
-                family.FamilyMembers = await _personRepository.GetAllByRecipientId(family.RowKey);
+                var familyMembers = await _personRepository.GetAllByRecipientId(family.RowKey);
+                family.FamilyMembers.AddRange(familyMembers); 
             }
             return recipients;
         }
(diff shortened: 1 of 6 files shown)


GiEnJul/Controllers/AdminController.cs
```diff
@@ -42,8 +42,14 @@ public async Task<IEnumerable<Giver>> GetGiversAsync()
...
         [HttpGet("recipients")]
-        public async Task<IEnumerable<Recipient>> GetRecipientsAsync() {
-            return await _recipientRepository.GetAllAsModelAsync();
+        public async Task<List<Recipient>> GetRecipientsAsync() {
+            var recipients = await _recipientRepository.GetAllAsModelAsync();
+            foreach (var family in recipients)
+            {
+                var familyMembers = await _personRepository.GetAllByRecipientId(family.RowKey);
+                family.FamilyMembers.AddRange(familyMembers); 
+            }
```
Sommerli, line 51: Small suggestion to change "family" to "recipient" in foreach, and assign the list as the new value instead of add to the empty list.
```suggestion
            foreach (var recipient in recipients)
            {
                recipient.FamilyMembers = await _personRepository.GetAllByRecipientId(recipient.RowKey);
            }
```
This could also be done with List.ForEach method combined with LINQ, but i do not know if it is equally fast or not. https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.foreach?view=netcore-3.1

Sommerli (changes requested):
Small suggestion to AdminController

GiEnJul/Controllers/AdminController.cs
elisabethnetland: Changes implemented. Did not change to List.ForEach though, I think it looks more readable as it is :)

Sommerli (approved):

Issue #209, opened 2021-07-13
Fix naming and warnings


TheseusDeus:
After #203 we get 3 warnings in the console, one for the naming of attribute "justify" on AdminPage.tsx:91, this should be "justifyContent".
Datatable and DatatableRecipient should have better naming schemes
in DataTableRecipient:50 where each element is generated we get a warning. Each child should have a unique key. Something like <Accordion key={"recipient"+x.referenceid}>
......
See this for more info: https://fb.me/react-warning-keys

PR #213, opened 2021-07-13
Display all familymembers

### LINKED ISSUE #210: Frontend - Display persons in each family in overview page

elisabethnetland:
Closes #210 & #209 
Changes: 
- Changed to more appropriate naming of files. 
- Displays persons for each recipient in the RecipientTable.

Alert:
The changes in the backend are old, don't know why they ended up in this PR.


TheseusDeus (changes requested):
Looks great, however:
Take a look at #209. You have fixed the naming scheme, but errors from that issue is still present
And there are more errors in the console now as well. 
When using data.map(x) => ( every child should have a unique key based on what kind of element this is.
When adding attributes to MaterialUI or other elements check what the correct name should be. Most of them are easy fixes, like justify -> justifyContent and rowsMax -> maxRows

TheseusDeus (approved):

GiEnJul/ClientApp/src/components/admin/common/RecipientTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const handleMatched = (input: Boolean) => {
+    if (input) {
+      return classes.matched;
+    } else {
+      return classes.notMatched;
+    }
+  };
+
+  const handleGender = (gender: Number, age: Number) => {
```
joffe97, line 48: The gender type in common/enums/Genders.tsx should be used here instead of Number. This applies to other places as well.

GiEnJul/ClientApp/src/components/admin/common/RecipientTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  return (
+    <Container>
+      {data.map((recipient) => (
+        <Accordion 
+        key={recipient.partitionKey}>
+          <AccordionSummary
+            expandIcon={<ExpandMore />}
+            aria-controls="panel1bh-content"
+            id="panel1bh-header"
```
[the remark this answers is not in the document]
joffe97: The partition_key or row_key + "something" can also be used here instead of the index.

GiEnJul/ClientApp/src/components/admin/common/RecipientTable.tsx
```diff
@@ -0,0 +1,134 @@
...
+              <Typography>Ønsker:</Typography></AccordionDetails>
+            {recipient.familyMembers.map((person) => (
+              <div
+              key={person.rowKey}>
+              <AccordionDetails >
+              <Typography className={classes.smallColumn}> 
+                {handleGender(person.gender, person.age)} 
+              </Typography>
+              <Typography className={classes.smallColumn}> {person.age} år </Typography>
+              <Typography className={classes.largeColumn}> {person.wish} </Typography>
```
joffe97, line 106: Does this handle cases with "age adjusted" gifts?

GiEnJul/ClientApp/src/components/admin/overview/Recipient.tsx
```diff
@@ -2,17 +2,24 @@ import { Container, Grid, TextField } from "@material-ui/core";
 import { Search } from "@material-ui/icons";
 import * as React from "react";
 import { useEffect, useState } from "react";
-import DatatableRecipient from "../common/DataTableRecipient";
+import Datatable from "../common/RecipientTable";
 import { GiverType } from "./Giver";
 
+export interface PersonType {
```
joffe97, line 8: This is very similar to the PersonType in InstitutionForm.tsx. I think it the PersonType should be placed into common, so the same type can be used both places. I don't think that the partitionKey and rowKey is necessary. If they are you could extend the type in the common folder.

joffe97 (changes requested):

PR #225, opened 2021-07-14
Post connection in admin controller

### LINKED ISSUE #224: Backend - POST matching of giver and recipient
Create a function with a giver and recipient as inputs in adminController. 
This function uses the ConnectionsRepo, RecipientRepo and GiverRepo to insert the connection in the entities.

elisabethnetland:
Closes #224 & #230
Added: 
- HttpPost method in adminController
- Constructer in ConnectionEntity
- insert connection method from recipientKey and giverKey in connectionRepo
- delete connection in connectionRepo if error

### DIFF
--- a/GiEnJul/Entities/Connection.cs
+++ b/GiEnJul/Entities/Connection.cs
@@ -7,13 +7,41 @@
 
         public Connection(string receiverLocation, string rowKey) : base(receiverLocation, rowKey)
         { }
+        public Connection(Giver giver, Recipient recipient)
+        { 
+            if(giver.PartitionKey != recipient.PartitionKey){
+                throw new System.Exception();
+            }
+            RowKey = $"{recipient.RowKey}_{giver.RowKey}"; 
+            PartitionKey = giver.PartitionKey;
+
+            Dinner = recipient.Dinner;
+            Dessert = recipient.Dessert;
+            Note = recipient.Note;
+            EventName = recipient.EventName;
+            PersonCount = recipient.PersonCount;
+            ReceiverLocation = recipient.Location;
+
+            SubmitterFullName = giver.FullName;
+            SubmitterEmail = giver.Email;
+            SubmitterPhoneNumber = giver.PhoneNumber;
+
+            Institution = recipient.Institution;
+            ReferenceId = recipient.ReferenceId;
+
+            MaxReceivers = giver.MaxRecievers;
+            GiverLocation = giver.Location;
+            GiverFullName = giver.FullName;
+            GiverEmail = giver.Email;
+            GiverPhoneNumber = giver.PhoneNumber;
+        }
 
         // Family info
         public string Dinner { get; set; }
         public string Dessert { get; set; }
         public string Note { get; set; }
-        public string Event { get; set; }
-        public string PersonCount { get; set; }
+        public string EventName { get; set; }
+        public int PersonCount { get; set; }
         public string ReceiverLocation { get; set; }
 
         // Submitter contact info
@@ -26,7 +54,7 @@
         public string ReferenceId { get; set; }
 
         // Giver info
-        public string MaxReceivers { get; set; }
+        public int MaxReceivers { get; set; }
         public string GiverLocation { get; set; }
 
         // Giver contact info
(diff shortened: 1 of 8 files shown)


GiEnJul/Entities/Connection.cs
```diff
@@ -7,13 +7,41 @@ public Connection()
 
         public Connection(string receiverLocation, string rowKey) : base(receiverLocation, rowKey)
         { }
+        public Connection(Giver giver, Recipient recipient)
+        { 
+            if(giver.PartitionKey != recipient.PartitionKey){
+                throw new System.Exception();
```
RonHaland, line 13: It's good to think about what kind of exception you are throwing here, so that the application can handle it the proper way.
This might be a good time for creating a custom exception, maybe an InvalidEventException or InvalidLocationException with the message ("Locations/events must match for giver and recipient")?

GiEnJul/Features/ConnectionRepository.cs
```diff
@@ -4,27 +4,27 @@
...
 
-
 namespace GiEnJul.Features
 {
     public interface IConnectionRepository
     {
-        Task<Connection> InsertOrReplaceAsync(Models.Giver giver);
-        Task<Connection> InsertOrReplaceAsync(Models.Recipient recipient);
+        Task<Connection> InsertOrReplaceAsync(Models.Giver giver, Models.Recipient recipient);
+        Task<Connection> DeleteConnectionAsync(Entities.Connection connection);
```
RonHaland, line 12: If we want to stay consistent, these should return a connection model, I assume this is the connection entity.

GiEnJul/Entities/Connection.cs
elisabethnetland: I agree! I would suggest to create an issue for this task and keep it separate from this PR. What do you think?

GiEnJul/Features/ConnectionRepository.cs
elisabethnetland: @Sommerli do you have any comments about this?

GiEnJul/Entities/Connection.cs
RonHaland: Could do that, but I just really dislike merging this as it is 😨

GiEnJul/Infrastructure/AutoMapperProfile.cs
```diff
@@ -31,9 +31,8 @@ public AutoMapperProfile()
                 .ForMember(dest => dest.MatchedGiver, act => act.Ignore());
 
             CreateMap<Models.Recipient, Entities.Recipient>()
-                .ConstructUsing(x => new Entities.Recipient(x.Location, x.EventName))
-                .ForMember(dest => dest.RowKey, act => act.Ignore())
-                .ForMember(dest => dest.PartitionKey, act => act.Ignore())
+                .ForMember(dest => dest.RowKey, opt => opt.Condition(src => (!string.IsNullOrEmpty(src.RowKey))))
+                .ForMember(dest => dest.PartitionKey, opt => opt.Condition(src => (!string.IsNullOrEmpty(src.PartitionKey))))
```
RonHaland, line 35: I have never used opt.Condition() before, what happens if one of these values are empty?

RonHaland (approved):

GiEnJul/Infrastructure/AutoMapperProfile.cs
elisabethnetland: It was added to prevent the mapping from creating a new entity when an entity is updated. So if the field is empty, then a new rowKey and partitionKey is created, if not it maps. Makes sense?

GiEnJul/Features/ConnectionRepository.cs
Sommerli: @RonHaland is correct, @elisabethnetland. Return types should be Models.Connection

Sommerli: As our project does not have a Models.Connection-class, the return type could be the Rowkey and Partitionkey of the created entity. This way we still have a reference to the Connection entity, without needing to handle entities in the Controller.

The current usage of the returned Connection is for keeping in case a exception is thrown, and the connection needs to be deleted. DeleteAsync could be replaced to receive a RK and a PK of the row to delete. 
image

@RonHaland

PR #229, opened 2021-07-14
Put states into object states

### LINKED ISSUE #228: Put states in institution/FormPerson.tsx into state objects.

joffe97:
Fix #228

### DIFF
--- a/GiEnJul/ClientApp/src/components/institution/FormPerson.tsx
+++ b/GiEnJul/ClientApp/src/components/institution/FormPerson.tsx
@@ -48,10 +48,8 @@ const InstitutionPerson: FC<PersonProps> = ({
   const getFormDataSetter = (target: keyof typeof formDataState) => {
     return (value: unknown) => {
       setFormDataState((prev) => {
-        {
-          prev[target] = value;
-          return { ...prev };
-        }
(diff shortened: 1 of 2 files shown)


GiEnJul/ClientApp/src/components/institution/FormPerson.tsx
```diff
@@ -16,132 +16,164 @@ import { isNotNull } from "../InputFields/Validators/Validators";
...
+  const [formDataState, setFormDataState] = useState(initFormDataState);
+  const [validFormState, setValidFormState] = useState(initValidFormState);
+
+  const getFormDataSetter = (target: keyof typeof formDataState) => {
+    return (value: unknown) => {
+      setFormDataState((prev) => {
+        {
+          prev[target] = value;
+          return { ...prev };
+        }
```
Sommerli, line 54:
```suggestion
          prev[target] = value;
          return { ...prev };
```
Could this bracket pair be removed?

joffe97: Yes it could! I see that this applies to other places as well.

TheseusDeus (approved):

PR #250, opened 2021-07-19
Task231 frontend connections

elisabethnetland:
The adminOverview is now structured this way: 
Adminpage is the main macro component that handles authorization. 
The adminTab handles the switches between overview, suggested matches & finished matches. 
In the overview-folder:
- OverviewMacrocomponent fetches the data from giver and recipient and stores the state of the selected giver and recipient for matching. It has three states, handles the changes in selected giver/recipient and posts a connection. It sends the data to giver and recipient as well as how to handle change in the selected giver/recipient as props. 
- Giver/Recipient handles searching and this is where we can implement filtering as well. It passes the same props as OverviewMacro to the dataTables. 
In the common-folder: 
- Recipient/GiverTable fomats the data in a presentable way. 

Edit: 
Have created a new issue for better user Experience #262 when connecting a recipient and giver.


GiEnJul/ClientApp/src/components/admin/common/RecipientTable.tsx
```diff
@@ -37,14 +44,6 @@ const DatatableRecipient: React.FC<Props> = ({ data }) => {
...
 
-  const handleMatched = (input: Boolean) => {
-    if (input) {
-      return classes.matched;
-    } else {
-      return classes.notMatched;
-    }
-  };
-
   const handleGender = (gender: Number, age: Number) => {
```
joffe97, line 47: There is a gender enum in `common/enums/Gender.tsx` that can be used instead of numbers. This will improve readability. 

I think the mapping to string should be done using a common function, if possible. There is one in `common/constants/Genders.tsx` that has an association between the Gender enum and strings. 
This doesn't map based on age though. I don't think this is necessary though, as this is only for admins. The `constants/Genders.tsx` could maybe be converted into a function that takes Gender and age as arguments, although I think this won't be needed.

GiEnJul/ClientApp/src/components/admin/common/RecipientTable.tsx
```diff
@@ -72,7 +71,20 @@ const DatatableRecipient: React.FC<Props> = ({ data }) => {
   return (
     <Container>
       {data.map((recipient, index) => (
-        <Accordion key={`acc_recipient_${index}`}>
+        <Accordion
+          onChange={() =>
+            handleRecipientChange(recipient.rowKey, recipient.partitionKey)
+          }
+          key={`acc_recipient_${index}`}
+          style={
```
joffe97, line 79: Not sure if this should be inline CSS or not. Could probably be added to a hook and used as a class instead?

GiEnJul/ClientApp/src/components/admin/overview/Giver.tsx
```diff
@@ -3,42 +3,19 @@ import { Search } from "@material-ui/icons";
...
-        .then((json) => setData(json))
-        .catch((errorStack) => {
-          console.log(errorStack);
-        });
-    }
-    fetchGivers();
-  }, []);
-
   const search = (input: GiverType[] | []) => {
     const keys = input[0] && Object.keys(input[0]);
```
joffe97, line 20: Nice and clean! Didn't know about the Object.keys() function 😮

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
```diff
@@ -0,0 +1,191 @@
...
+import usePost from "../../../hooks/usePost";
+
+export interface SelectedConnectionType {
+  giverRowKey: string;
+  giverPartitionKey: string;
+  recipientRowKey: string;
+  recipientPartitionKey: string;
+}
+
+export interface GiverType {
```
joffe97, line 14: Would it be nice to have the `GiverType`, `RecipientType` and `PersonType` in a `types.tsx` (or something) file?

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
```diff
@@ -0,0 +1,191 @@
...
+    useState<SelectedConnectionType>({
+      giverRowKey: "",
+      giverPartitionKey: "",
+      recipientRowKey: "",
+      recipientPartitionKey: "",
+    });
+  const [giverData, setGiverData] = useState<GiverType[] | []>([]);
+  const [recipientData, setRecipientData] = useState<RecipientType[] | []>([]);
+
+  async function fetchGivers() {
```
joffe97, line 67: The `fetchGivers` and `fetchRecipients` should be changed into using the `useFetch` function instead.

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
```diff
@@ -0,0 +1,191 @@
...
+    newGiverPartitionKey: string
+  ) => {
+    setSelectedConnection((prev) => ({ ...prev, giverRowKey: newGiverRowKey }));
+    setSelectedConnection((prev) => ({
+      ...prev,
+      giverPartitionKey: newGiverPartitionKey,
+    }));
+    setGiverData(
+      giverData.map((item) =>
+        item.rowKey === newGiverRowKey && item.isSuggestedMatch === false
```
joffe97, line 110: There should be a key attribute added here. This could be made by using e.g. the index and a custom string.

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
```diff
@@ -0,0 +1,191 @@
...
+          ? { ...item, isSelected: true }
+          : { ...item, isSelected: false }
+      )
```
joffe97, line 113: This can be shortended:
```suggestion
      giverData.map((item) =>
        ({ ...item, isSelected: (item.rowKey === newGiverRowKey && item.isSuggestedMatch === false)})
      )
```

joffe97: This applies below as well on line 129 `setRecipientData()`

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
```diff
@@ -0,0 +1,191 @@
...
+        item.rowKey === newRecipientRowKey && item.isSuggestedMatch === false
+          ? { ...item, isSelected: true }
+          : { ...item, isSelected: false }
+      )
+    );
+  };
+
+  const connectGiverRecipient = async () => {
+    console.log(selectedConnection.giverPartitionKey);
+    await fetch("api/admin", {
```
joffe97, line 140: Could the `usePost` hook be used here?

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
```diff
@@ -0,0 +1,191 @@
...
+      },
+      body: JSON.stringify({
+        GiverRowKey: selectedConnection.giverRowKey,
+        GiverPartitionKey: selectedConnection.giverPartitionKey,
+        RecipientRowKey: selectedConnection.recipientRowKey,
+        RecipientPartitionKey: selectedConnection.recipientPartitionKey,
+      }),
+    })
+      .then((response) => {
+        if (response.status === 201) {
```
joffe97, line 153: Is this added for later usage? If so, it would be nice to add a comment about it.

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
```diff
@@ -0,0 +1,191 @@
...
+        }
+      })
+      .catch((errorStack) => {
+        console.log(errorStack);
```
joffe97, line 157: I think the `console.error()` function would be better here?

GiEnJul/ClientApp/src/components/admin/overview/Recipient.tsx
```diff
@@ -3,55 +3,19 @@ import { Search } from "@material-ui/icons";
...
+  handleRecipientChange: (
+    newRecipientRowKey: string,
+    newRecipientPartitionKey: string
+  ) => void;
+};
 
-const Recipient = () => {
-  const [data, setData] = useState<RecipientType[] | []>([]);
+const Recipient: React.FC<Props> = ({ data, handleRecipientChange }) => {
   const [q, setQ] = useState("");
```
joffe97, line 17: I think there should be a more descriptive name here. E.g. `query`?

GiEnJul/Controllers/AdminController.cs
```diff
@@ -47,25 +49,27 @@ public async Task<IEnumerable<Giver>> GetGiversAsync()
...
+        public async Task<ActionResult> PostConnectionAsyc([FromBody] PostConnectionDto connectionDto)
+        {
+            var giver = await _giverRepository.GetGiverAsync(connectionDto.GiverPartitionKey, connectionDto.GiverRowKey);
+            var recipient = await _recipientRepository.GetRecipientAsync(connectionDto.RecipientPartitionKey, connectionDto.RecipientRowKey);
             recipient.FamilyMembers = await _personRepository.GetAllByRecipientId(recipient.RowKey);
-            var giver = await _giverRepository.GetGiverAsync(partitonKey, giverRowKey);
             var connection = await _connectionRepository.InsertOrReplaceAsync(giver, recipient);
             giver.IsSuggestedMatch = true;
-            giver.MatchedRecipient = recipientRowKey;
+            giver.MatchedRecipient = connectionDto.RecipientRowKey;
```
joffe97, line 70: Could this maybe be done with automapper?

joffe97 (changes requested):
Nice work! Added some nitpick comments.

GiEnJul/ClientApp/src/components/admin/common/RecipientTable.tsx
elisabethnetland: I think we should include the difference between gutt/mann even though is for admin. We will have to do it this way in the email we send to the givers so why not keep it consistent.

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
elisabethnetland: It is not working properly atm

GiEnJul/ClientApp/src/components/admin/overview/Recipient.tsx
elisabethnetland: agree

GiEnJul/ClientApp/src/components/admin/common/RecipientTable.tsx
joffe97: That's true. Would it be better to change the `/constants/Genders.tsx` into a function then that uses gender and age as arguments? If age is undefined "mann"/"kvinne" could be returned. I think this would be nice for clarity and reusability.
This may be a rather big task though and could be put into its own issue.

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
elisabethnetland: not working properly atm

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
elisabethnetland: agree

GiEnJul/ClientApp/src/components/admin/common/RecipientTable.tsx
elisabethnetland: This is work in progress. Will be separated in a new file

GiEnJul/Controllers/AdminController.cs
elisabethnetland: The connectionDto is not a model input, it is only four strings so therefore it can not be mapped

GiEnJul/ClientApp/src/components/admin/common/RecipientTable.tsx
elisabethnetland: Gender is implemented. We can create an issue for creating a functions since we have to change other parts of the code when doing this, such as institutionForm.

GiEnJul/ClientApp/src/components/admin/common/GiverTable.tsx
```diff
@@ -37,18 +41,21 @@ const Datatable: React.FC<Props> = ({ data }) => {
...
       {data.map((giver, index) => (
-        <Accordion key={`acc_giver_${index}`}>
+        <Accordion
+          onChange={() => handleGiverChange(giver.rowKey, giver.partitionKey)}
+          key={`acc_giver_${index}`}
+          style={
+            giver.isSelected
+              ? {
+                  background:
+                    "linear-gradient(45deg, #D6F0EB 30%, #E2FFF9 90%)",
```
TheseusDeus, line 54: do we want direct styling in our components?

TheseusDeus, line 54: Do we want styling directly in our components?

elisabethnetland: Written comments in the code so that we change this when implementing styling

joffe97 (approved):
Nice work! Can you make an issue on the optimization problem?

Issue #218, opened 2021-07-13
Backend - delete giver in admin controller


elisabethnetland:
Create a function in admin controller that takes the partitionkey of a giver as input and deletes this entity from TableStorage.

PR #264, opened 2021-07-21
Deletegiverfunc

TheseusDeus:
Fixes issue #218 

DELETE api/admin/giver with a DeleteGiverDto as the body will now delete that Giver from the DB.
DeleteGiverDto consists of a PK and RK for the Giver in question

### DIFF
--- a/GiEnJul/Controllers/AdminController.cs
+++ b/GiEnJul/Controllers/AdminController.cs
@@ -45,7 +45,7 @@ namespace GiEnJul.Controllers
         // }
 
         [HttpGet("givers")]
-        // [Authorize(Policy = "ReadGivers")] This will deny any unauthorized requests, but will break the application atm
+        // [Authorize(Policy = "ReadGiver")] This will deny any unauthorized requests, but will break the application atm
         public async Task<IEnumerable<Giver>> GetGiversAsync()
         {
             return await _giverRepository.GetAllAsModelAsync();
@@ -92,6 +92,7 @@ namespace GiEnJul.Controllers
             return Ok();
         }
         [HttpDelete("giver")]
+        [Authorize(Policy = "DeleteGiver")]
         public async Task<ActionResult> DeleteGiverAsync([FromBody] DeleteGiverDto giverDto)
         {
             try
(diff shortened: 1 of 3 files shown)


GiEnJul/Controllers/AdminController.cs
```diff
@@ -89,5 +91,26 @@ public async Task<ActionResult> PostConnectionAsyc(string giverRowKey, string re
...
+                    return NotFound($"Giver with PK = {giverDto.PartitionKey} and RK = {giverDto.RowKey} not found");
+                var deletedGiver = await _giverRepository.DeleteAsync(giverToDelete);
+                return Ok();
+            }
+            catch (System.Exception err)
+            {
+                System.Console.WriteLine(err);
+                return StatusCode(StatusCodes.Status500InternalServerError,
+                "Error deleting data");
+            }
```
Sommerli, line 113: The behavior of this block is equal to the standard error handler. 
The default handler would write the error to terminal, and return a HTTP status 500. Because of this, i belive its not necessary with a try catch here.

Sommerli (commented):
Some comments on logging and try catch behaviour

GiEnJul/Controllers/AdminController.cs
```diff
@@ -89,5 +91,26 @@ public async Task<ActionResult> PostConnectionAsyc(string giverRowKey, string re
...
             return Ok();
         }
+        [HttpDelete("giver")]
+        public async Task<ActionResult> DeleteGiverAsync([FromBody] DeleteGiverDto giverDto)
+
+        {
+            System.Console.WriteLine("im here");
+            try
+            {
+                System.Console.WriteLine("Trying");
```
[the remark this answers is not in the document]
TheseusDeus: this was for debugging yeah

GiEnJul/Controllers/AdminController.cs
TheseusDeus: So I could remove everything regarding error handling then, and just have line 99, 102 and 103 @Sommerli ?

TheseusDeus: Fixed with 6971c59 @Sommerli

Sommerli (approved):

PR #273, opened 2021-07-22
Renamed features to repositories

Sommerli:
Recommend using following settings when reviewing this
image


GiEnJul/Controllers/AdminController.cs
```diff
@@ -9,6 +8,7 @@
 using GiEnJul.Models;
 using GiEnJul.Dtos;
```
TheseusDeus, line 9: The duplicate import should be removed with 9c1d949, may need to merge first

TheseusDeus (commented):
Likes the renaming. Makes it way more clear

GiEnJul/Controllers/AdminController.cs
Sommerli: Merge conflicts are now resolved

TheseusDeus (approved):

RonHaland (approved):

PR #278, opened 2021-07-23
Fixed bugs on institution registration page

### LINKED ISSUE #253: Frontend - Cannot sumbit institution registration form if "alderstilpasset gave" is not selected.

joffe97:
Fix #253 

Fixed multiple bugs:
- Errorfields didn't reset (in some cases) after submitting.
- Persons states got mixed up.
- Recursion error when updating users (sometimes).
- Wishinput field didn't update parent component.

### DIFF
--- a/GiEnJul/ClientApp/src/components/institution/FormFood.tsx
+++ b/GiEnJul/ClientApp/src/components/institution/FormFood.tsx
@@ -49,7 +49,7 @@ const FormDinner: FC<IFormDinner> = ({
   };
 
   useEffect(() => {
-    if (viewErrorTrigger) setViewError(true);
+    setViewError(!!viewErrorTrigger);
   }, [viewErrorTrigger]);
 
   useEffect(() => {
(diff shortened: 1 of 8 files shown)


GiEnJul/ClientApp/src/components/institution/FormFood.tsx
```diff
@@ -49,7 +49,7 @@ const FormDinner: FC<IFormDinner> = ({
   };
 
   useEffect(() => {
-    if (viewErrorTrigger) setViewError(true);
+    setViewError(!!viewErrorTrigger);
```
TheseusDeus, line 52: why are you using double "!" here?

joffe97: This is to convert from int to boolean. E.g. converts first to false then true.

TheseusDeus: aah thats JS for you

TheseusDeus (approved):

PR #285, opened 2021-07-23
Endpoint to fetch a randomly selected list of givers that are ready to be assigned a family

Sommerli:
Created api-function to fetch a specified quantity of Givers on a location, that could be suggested for connections.

Call this endpoint:
https://localhost:5001/api/admin/Suggestions/Giver for single Giver per familysize
https://localhost:5001/api/admin/Suggestions/Giver/x for x Givers per familysize (or all if there is no more than x) 

Results from three different calls.
image


GiEnJul/Controllers/AdminController.cs
```diff
@@ -138,5 +138,17 @@ public async Task<ActionResult> DeleteGiverAsync([FromBody] DeleteGiverDto giver
...
+
+        [HttpGet("Suggestions/Giver/{quantity}")]
+        [HttpGet("Suggestions/Giver")]
+        public async Task<IList<GiverDataTableDto>> GetSuggestedGiversAsync([FromBody] string location, int quantity = 1)
+        {
+            if (quantity < 1) throw new ArgumentOutOfRangeException();
+
+            var activeEvent = await _eventRepository.GetActiveEventForLocationAsync(location);
+            var suggestions = await _giverRepository.GetSuggestionsAsync(activeEvent, location, quantity);
+
```
Sommerli, line 150:
```suggestion
var givers = await _giverRepository.GetSuggestionsAsync(activeEvent, location, quantity);
var suggestions = SuggestionHelper.GetSuggestions(givers, quantity);
```

Alternative is to call Helper-function here. To keep the repository to SOLID principles.

PR #289, opened 2021-07-26
Added endpoint to fetch random set of Givers that are unmatched 

Sommerli:
Added Helper-class to pick a random set of Givers/Receivers.

Added PersonCount to Models.Recipient

Added api-function to fetch random set of Givers based of location, and an amount.


TheseusDeus (changes requested):
The ClientApp/package-lock.json file should not be present in the PR, especially since you have not done anything about the frontend.
The lockfileVersion is also at "1", while the one in master is at "2"

Sommerli:
> The ClientApp/package-lock.json file should not be present in the PR, especially since you have not done anything about the frontend.
> The lockfileVersion is also at "1", while the one in master is at "2"

Removed changes from package-lock now

TheseusDeus (approved):

PR #293, opened 2021-07-27
Added api-function for fetching a random set of Recipient based of location and event

Sommerli:
Added helper for creation of filters that are queried multiple times.

Added function to fetch Recipients based of location.


RonHaland (approved):
Some say it might be too good 🤨

Good job 🤠

PR #295, opened 2021-07-27
Selectform bug

### LINKED ISSUE #287: Frontend - Bug on mobile in the locations select, when registering giver
There is a bug in the location select on the giver registration page. This problem probably is in the SelectForm component. When locations is loaded on mobile, the first location is automatically selected (usually Bodø). This doesn't update the state of the chosen location though. Image below shows what happens when pressing continue. 
image
I would suggest to add an empty option for the SelectForm on mobile that is used as default. This option should disappear when a location is selected.

joffe97:
Fix #287
Fixed bug in selectform component on mobile, where it used a default value that didn't update in state.

### DIFF
--- a/GiEnJul/ClientApp/src/components/InputFields/SelectForm.tsx
+++ b/GiEnJul/ClientApp/src/components/InputFields/SelectForm.tsx
@@ -36,7 +36,7 @@ const SelectForm: FC<ISelect> = ({
 }) => {
   const [isMob, setIsMob] = useState(false);
 
-  const jsxOptions = useMemo(() => {
+  const jsxOptions = () => {
     if (isMob) {
       if (options && options.length !== 0) {
         return (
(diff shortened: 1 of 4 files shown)


Sommerli (approved):

GiEnJul/ClientApp/src/components/InputFields/SelectForm.tsx
```diff
@@ -45,37 +77,22 @@ const SelectForm: FC<ISelect> = ({
...
-                value={o.value}
-                className="text-capitalize"
-              >
-                {o.text}
-              </option>
-            ) : (
-              <MenuItem
-                key={`${name}_${o.text}`}
-                value={o.value}
-                className="text-capitalize"
```
elisabethnetland, line 73: Small thing but the "text-capitalize" makes the first letter in each word to upper case. Would suggest to remove line 65 and73.
image

Issue #297, opened 2021-07-27
Add location to admins


joffe97:
All admin should have a location assigned to them. This could be implemented with Auth0.

Issue #286, opened 2021-07-26
Backend - Add functionality for downloading excel sheet of recipients


joffe97:
Need input from Karianne on how the sheet should be designed and which columns should be included.

PR #298, opened 2021-07-27
Frontend delivery excel

joffe97:
Added button in completed suggestions to download the delivery excel sheet for given location.
The location is set to Nittedal as default. This must be changed after #297 is completed.
The button won't work before the backend part is completed in #286.


GiEnJul/ClientApp/src/components/admin/connections/completed/DeliveryExcelDownload.tsx
```diff
@@ -0,0 +1,20 @@
...
+import React, { useState } from "react";
+
+const DeliveryExcelDownload = () => {
+  const [location, _] = useState("Nittedal");
+  return (
+    <Button
+      variant="contained"
+      color="primary"
+      startIcon={<GetAppIcon />}
+      href={`/api/admin/excel/delivery/${location}`}
```
RonHaland, line 12: Two things to keep in mind here,
1. this url be different if we split frontend and backend
2. this endpoint should be require authorization.

Could button use an onclick action that makes use of the ApiService class?

joffe97: Changed this into a onclick action that uses the ApiService class. The access token will be passed down later from the AdminPage macro.

TheseusDeus (approved):

PR #299, opened 2021-07-28
get LogLevel from config, correctly log exceptions

RonHaland:
Small fixes to how exceptions were logged.
Get minimum logging level from config, so that this can more easily be adjusted

### DIFF
--- a/GiEnJul/Repositories/GenericRepository.cs
+++ b/GiEnJul/Repositories/GenericRepository.cs
@@ -31,7 +31,7 @@ namespace GiEnJul.Repositories
             }
             catch (StorageException e)
             {
-                _log.Fatal("Could not create or connect to Azure Table Storage.\n{@0}", e.ToString());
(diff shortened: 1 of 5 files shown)


GiEnJul/Repositories/GenericRepository.cs
```diff
@@ -50,7 +50,7 @@ protected async Task<T> DeleteAsync(T entity)
             }
             catch (Exception e)
             {
-                _log.Error("Exception while trying to delete Entity:{@entity}, in table:{@tablename}. \n{@Exception}", entity, _table.Name, e);
+                _log.Error(e, "Exception while trying to delete Entity:{@entity}, in table:{@tablename}", entity, _table.Name);
```
RonHaland, line 53: To avoid getting the exception object mixed into the params[] for string formatting, we can pass it as the first argument in log.Error or log.Fatal.

joffe97 (approved):

elisabethnetland (approved):

PR #302, opened 2021-07-28
Fetchtoapiservice

### LINKED ISSUE #277: Frontend - Refactor usage of fetch to apiservice
- [x] ClientApp\src\common\constants\Locations.tsx(2): const locations = await fetch('/api/event/ActiveLocations')
- [x] ClientApp\src\components\admin\overview\OverviewMacro.tsx(20): await fetch("api/admin/givers", {
- [x] ClientApp\src\components\admin\overview\OverviewMacro.tsx(33): await fetch("./api/admin/recipients", {
- [x] ClientApp\src\components\admin\overview\OverviewMacro.tsx(79): await fetch("api/admin", {
- [x] ClientApp\src\components\institution\InstitutionForm.tsx(316): await fetch("/api/recipient", {
- [x] ClientApp\src\components\registerGiver\Summary.tsx(88): await fetch("api/giver", {
- [x] ClientApp\src\components\verifyConnection\verifyConnection.tsx(20): fetch(`verify/${giverRowKey}/${recipientRowKey}/${partitionKey}`, {

Example transition:

`  async function fetchGivers() {
    await fetch("api/admin/givers", {
      method: "GET", // *GET, POST, PUT, DELETE, etc.
      headers: {
        "Content-Type": "application/json",
      },
    })
      .then((response) => response.json())
      .then((json) => setGiverData(json))
      .catch((errorStack) => {
        console.error(errorStack);
      });
  }
`
to
`  async function fetchGiverData() {
    await apiservice.get("admin/givers").then((resp) => setGiverData(resp.data));
  }
`

TheseusDeus:
fixes #277 so that all instances of fetch is now on the form of apiservice

### DIFF
--- a/GiEnJul/ClientApp/src/components/verifyConnection/verifyConnection.tsx
+++ b/GiEnJul/ClientApp/src/components/verifyConnection/verifyConnection.tsx
@@ -20,7 +20,7 @@ const VerifyConnection: React.FC<VerifyConnection> = () => {
 
   useEffect(() => {
     apiservice
-      .postEmpty(`verify/${giverRowKey}/${recipientRowKey}/${partitionKey}`)
+      .post(`verify/${giverRowKey}/${recipientRowKey}/${partitionKey}`, {})
       .then((response) => setError(!(response.status == 200)));
   }, []);
 
--- a/GiEnJul/ClientApp/src/components/verifyConnection/verifyConnection.tsx
+++ b/GiEnJul/ClientApp/src/components/verifyConnection/verifyConnection.tsx
@@ -1,6 +1,7 @@
 ﻿import * as React from "react";
 import { useEffect, useState } from "react";
 import { RouteComponentProps, useParams } from "react-router";
+import ApiService from "../../common/functions/apiServiceClass";
 
 interface RouteParameters {
   giverRowKey: string;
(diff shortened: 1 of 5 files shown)


Sommerli (approved):

GiEnJul/ClientApp/src/components/verifyConnection/verifyConnection.tsx
```diff
@@ -15,11 +16,12 @@ const VerifyConnection: React.FC<VerifyConnection> = () => {
...
 
   const [error, setError] = useState(false);
+  const apiservice = new ApiService();
 
   useEffect(() => {
-    fetch(`verify/${giverRowKey}/${recipientRowKey}/${partitionKey}`, {
-      method: "POST",
-    }).then((response) => setError(!response.ok));
+    apiservice
+      .post(`verify/${giverRowKey}/${recipientRowKey}/${partitionKey}`, {})
```
joffe97, line 23: The data attribute in the post method should probably be set to {} as default. This way it doesn't have to be set in every call that doesn't contain any data.

TheseusDeus: usually when doing  a post, there are some data attached, so it is better to just set it as empty when needed

PR #303, opened 2021-07-29
Fixed bug on Admin - Overview

elisabethnetland:
Changes: 
- It is now possible to search when typing upperCase
- Added checks to avoid connecting givers and recipients several times
- Removed coloring of giver & recipient when selecting since it will only be displayed when rendering

### DIFF
--- a/GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
+++ b/GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
@@ -6,10 +6,8 @@ import Recipient from "./Recipient";
 import { GiverType, RecipientType, SelectedConnectionType } from "./Types";
 
 const initState: SelectedConnectionType = {
-  giverRowKey: "",
-  giverPartitionKey: "",
-  recipientRowKey: "",
-  recipientPartitionKey: "",
+  giver: undefined,
+  recipient: undefined,
 };
 
 function OverviewMacro() {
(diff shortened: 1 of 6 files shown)


GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
```diff
@@ -72,11 +68,11 @@ function OverviewMacro() {
       .post(
         "admin/",
         JSON.stringify({
-          GiverRowKey: selectedConnection.current.giverRowKey,
-          GiverPartitionKey: selectedConnection.current.giverPartitionKey,
-          RecipientRowKey: selectedConnection.current.recipientRowKey,
+          GiverRowKey: selectedConnection.current.giver?.rowKey,
```
TheseusDeus, line 71: Is it possible to just stringify it with {selectedConnection.current.GiverType, ...RecipientType} or {selectedConnection}?

TheseusDeus (approved):

GiEnJul/ClientApp/src/components/admin/overview/OverviewMacro.tsx
TheseusDeus: Oh that was supposed to be a "comment", oh well

elisabethnetland: I tried to do it like that but it has to be declared somehow. Don't know why.

Issue #301, opened 2021-07-28
Frontend - NavBarAdmin


elisabethnetland:
Create Navbar for the admin page according with functional component and material ui. The Navbar for landing page is already made, and can be used for inspiration.

TheseusDeus:
done

PR #305, opened 2021-07-29
Refactoredadminpage

TheseusDeus:
The useUser() hook can now extract role information from auth0
Refactored the adminpage so that content is displayed depending on role, and error handling for this.
adminpage now passes the locationprop down to completed
Navbar for institutionpage is now only a log out button

### DIFF
--- a/GiEnJul/ClientApp/src/common/components/NavMenuAdmin.tsx
+++ b/GiEnJul/ClientApp/src/common/components/NavMenuAdmin.tsx
@@ -1,43 +1,90 @@
-import * as React from 'react';
-import { Collapse, Container, Navbar, NavbarBrand, NavbarToggler, NavItem, NavLink } from 'reactstrap';
-import { Link } from 'react-router-dom';
-import './NavMenuAdmin.css';
-import logo from './../../styling/img/logo_green.svg';
+import * as React from "react";
+import { Link } from "react-router-dom";
+import {
+  Collapse,
+  Container,
+  Navbar,
+  NavbarBrand,
+  NavbarToggler,
+  NavItem,
+  NavLink,
+} from "reactstrap";
+import LogOutButton from "../../components/login/LogOutButton";
+import logo from "./../../styling/img/logo_green.svg";
+import "./NavMenuAdmin.css";
 
-export default class NavMenuAdmin extends React.PureComponent<{}, { isOpen: boolean }> {
-    public state = {
-        isOpen: false
-    };
+export default class NavMenuAdmin extends React.PureComponent<
+  { role?: string },
+  { isOpen: boolean }
+> {
+  public state = {
+    isOpen: false,
+  };
 
-    public render() {
-        return (
-                <Navbar className="navbar-expand-sm navbar-toggleable-sm border-bottom box-shadow" light>
-                    <Container>
-                        <NavbarBrand tag={Link} to="/">
-                            <img className={"logo-small"} src={logo}></img>
-                        </NavbarBrand>
-                        <NavbarToggler onClick={this.toggle} className="mr-2"/>
-                        <Collapse className="d-sm-inline-flex flex-sm-row-reverse" isOpen={this.state.isOpen} navbar>
-                            <ul className="navbar-nav flex-grow">
-                                <NavItem>
-                                    <NavLink tag={Link} className="text-dark" to="/admin">Admin</NavLink>
-                                </NavItem>
-                                <NavItem>
-                                    <NavLink tag={Link} className="text-dark" to="/registrer-familie">Registrer familie</NavLink>
-                                </NavItem>
-                                <NavItem>
-                                    <NavLink tag={Link} className="text-dark" to="/rediger">Rediger nettside</NavLink>
-                                </NavItem>
-                            </ul>
-                        </Collapse>
-                    </Container>
-                </Navbar>
-        );
+  public render() {
+    if (this.props.role == "Institution") {
+      return (
+        <Navbar
+          className="navbar-expand-sm navbar-toggleable-sm border-bottom box-shadow"
+          light
+        >
+          <Container>
+            <NavbarBrand tag={Link} to="/">
+              <img className={"logo-small"} src={logo}></img>
+            </NavbarBrand>
+            <NavbarToggler onClick={this.toggle} className="mr-2" />
+            <LogOutButton />
+          </Container>
+        </Navbar>
+      );
+    } else {
+      return (
+        <Navbar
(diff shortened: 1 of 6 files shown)


GiEnJul/ClientApp/src/common/components/NavMenuAdmin.tsx
```diff
@@ -1,43 +1,90 @@
...
+        <Navbar
+          className="navbar-expand-sm navbar-toggleable-sm border-bottom box-shadow"
+          light
+        >
+          <Container>
+            <NavbarBrand tag={Link} to="/">
+              <img className={"logo-small"} src={logo}></img>
+            </NavbarBrand>
+            <NavbarToggler onClick={this.toggle} className="mr-2" />
+            <Collapse
```
joffe97, line 51: This collapse tag should probably have a check that only renders it when `this.props.role === "Institution"`. This way the conditional render in line 25-39 could be removed.

GiEnJul/ClientApp/src/common/components/NavMenuAdmin.tsx
```diff
@@ -1,43 +1,90 @@
...
 
-    private toggle = () => {
-        this.setState({
-            isOpen: !this.state.isOpen
-        });
-    }
+  private toggle = () => {
+    this.setState({
+      isOpen: !this.state.isOpen,
+    });
```
joffe97, line 88: I think the prev state should be used here to avoid state bugs.
```suggestion
    this.setState((prev) => ({
      isOpen: !prev.isOpen,
    }));
```

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
```diff
@@ -21,15 +24,45 @@ function AdminPage() {
...
+            ErrorText={
+              "Du har ikke blitt tildelt en rolle enda, tilkall din GiEnJul admin"
+            }
+          />
+        );
+      default:
+        return <ErrorPage ErrorText={"En Feil har skjedd"} ErrorCode={500} />;
+    }
   } else {
+    console.log(role, location);
```
joffe97, line 64: Is this supposed to be here?

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
```diff
@@ -21,15 +24,45 @@ function AdminPage() {
...
-        <LogOutButton />
-        <AdminTab accessToken={userAccessToken} />
-      </>
+      <ErrorPage
+        ErrorText={
+          "Du har ikke blitt tildelt en lokasjon enda, tilkall din GiEnJul admin"
+        }
+      />
     );
+  } else if (role !== undefined && location !== undefined) {
```
joffe97, line 35: This `else if` should probably switch place with the `else` on line 63? This if sentence can grow larger if more conditional checks is added. I think that the checks should be ran in if/else if conditions first and if the code doesn't enter any of them, the wanted page will be rendered.

Could it be an idea to change the logic for which page should be rendered into another function, and use a switch case here instead?

joffe97 (changes requested):

GiEnJul/ClientApp/src/common/components/NavMenuAdmin.tsx
TheseusDeus: Did not think too much about how the navmenuadmin.tsx file looked cause of issue #301. Didn't want the PR to get too big

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
TheseusDeus: nah I'll remove it

GiEnJul/ClientApp/src/common/components/NavMenuAdmin.tsx
joffe97: Oh I agree. This should probably be done in #301 then.

joffe97: Hmm yeah it probably does.

GiEnJul/ClientApp/src/components/admin/AdminPage.tsx
TheseusDeus: It is not that much code, and not sure it's better to extract to another file. 
what do you mean by switching places with 35 and 63, then the else with <LoadingPage/> would always happen.
how would you like it?
if
else
  if
  if
I don't get it

joffe97: I think it would be a bit cleaner if we made a function e.g. `getPage`:
```typescript
const getPage = () => {
  if (location === null) return Page.NoLocation;
  else if (role === undefined || location === undefined) return Page.Loading;
  else if (role === "Admin") return Page.Admin;
  ...
}
```
The rendering could be changed to this then:
```typescript
switch (getPage()) {
  case Page.NoLocation:
    return <ErrorPage .../>;
  case Page.Loading:
    return <LoadingPage/>;
  case ...
}
```

TheseusDeus: That would be very clean, I'll make that helper function. It has to be in the same file to have access to the variables tho

joffe97: Yeah that would be nice 😄

TheseusDeus:
```typescript 
function getPage() {
    if (location === null) return "NoLocation";
    else if (role === undefined || location === undefined) return "Loading";
    else if (role === "Unspecified") return "Unspecified";
    else if (role === "Admin") return "Admin";
    else if (role === "Institution") return "Institution";
  }
```
....
```typescript
switch (getPage()) {
...
case "Admin":
      return (
        <>
          <NavMenuAdmin />
          <AdminTab accessToken={userAccessToken} location={location} />
        </>
      );
```
The ts compiler kills me since it says that the string could be undefined

TheseusDeus: The location variable can't be undefined since that would return the loading page. So added a {/*// @ts-ignore */} above the component and pushed it with f7aa4bf
have a look at it @joffe97

joffe97: Yeah looks like a flaw with typescript, because location will never be undefined here. You could probably change it into this: 
```typescript
<AdminTab accessToken={userAccessToken} location={location ?? ""} />
```

joffe97 (approved):
Looks nice!

PR #308, opened 2021-07-29
Backend changes to Suggestion-endpoint

Sommerli:
Added dto to use in frontend datatable

changed frombody to fromquery.


elisabethnetland (approved):
Should we use this dto for overview as well? And the difference between this method and the method used in overview is that here you only fetch some field and not all, right?

Sommerli:
> Should we use this dto for overview as well? And the difference between this method and the method used in overview is that here you only fetch some field and not all, right?
Probably, but it will not be implemented in the current issue.

PR #327, opened 2021-08-03
getConnectionsbyLocation

TheseusDeus:
added endpoint on admincontroller to get connections by location

### DIFF
--- a/GiEnJul/Controllers/AdminController.cs
+++ b/GiEnJul/Controllers/AdminController.cs
@@ -84,6 +84,15 @@ namespace GiEnJul.Controllers
             using var wb = ExcelGenerator.Generate(_mapper.Map<IEnumerable<DeliveryExcel>>(connections));
             return wb.Deliver("leveranse_liste.xlsx");
         }
+        [HttpGet("completed/{location}")]
+        [Authorize(Policy = "ReadConnection")]
+        public async Task<IEnumerable<Entities.Connection>> GetConnectionsByLocation(string location)
+        {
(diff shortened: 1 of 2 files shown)


GiEnJul/Controllers/AdminController.cs
```diff
@@ -84,6 +84,15 @@ public async Task<FileStreamResult> DownloadExcelDeliveryLocationAsync(string lo
             using var wb = ExcelGenerator.Generate(_mapper.Map<IEnumerable<DeliveryExcel>>(connections));
             return wb.Deliver("leveranse_liste.xlsx");
         }
+        [HttpGet("completed/{location}")]
+        [Authorize(Policy = "ReadConnection")]
+        public async Task<IEnumerable<Entities.Connection>> GetConnectionsByLocation(string location)
```
Sommerli, line 89: Controllers should not handle Entities. The repository function should return a Model-object, which then could create a DTO of the data you want to show.

GiEnJul/Controllers/AdminController.cs
```diff
@@ -84,6 +84,15 @@ public async Task<FileStreamResult> DownloadExcelDeliveryLocationAsync(string lo
...
+        {
+            var eventName = await _eventRepository.GetActiveEventForLocationAsync(location);
+            var connections = await _connectionRepository.GetAllByLocationEventAsync(location, eventName);
```
Sommerli, line 92: Change the return value of this function :)

Sommerli (changes requested):

PR #341, opened 2021-08-03
Added some changes to Verify-Connection page. 

Sommerli:
Frontend:
Shows a Spinner while waiting for response, an Error message if request was unsuccessful, and a thank you message if request was successful

Backend:
Quits verifyConnection-operation early if the connection already exists.

### DIFF
--- a/GiEnJul/ClientApp/src/components/verifyConnection/verifyConnection.tsx
+++ b/GiEnJul/ClientApp/src/components/verifyConnection/verifyConnection.tsx
@@ -1,8 +1,9 @@
-﻿import { makeStyles } from "@material-ui/core";
+﻿import { makeStyles, Typography } from "@material-ui/core";
 import * as React from "react";
 import { useEffect, useState } from "react";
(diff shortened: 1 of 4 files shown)


GiEnJul/ClientApp/src/components/verifyConnection/verifyConnection.tsx
```diff
@@ -11,33 +13,52 @@ interface RouteParameters {
...
+      .then((response) => {
+        setSuccess(response.status == 200);
+      });
   }, []);
 
-  if (error)
+  if (success)
+    return (
+      <>
+        <h1 className={classes.midScreen}>Takk til deg!</h1>
```
joffe97, line 44: I think material-ui typography should be used here: https://material-ui.com/api/typography/. I think there is a prop that can be added here (e.g. `align="center"`) instead of the midScreen class.

joffe97 (changes requested):

joffe97 (approved):

PR #344, opened 2021-08-03
Frontend - statistics

### LINKED ISSUE #330: Frontend - Statistics in overview

elisabethnetland:
Closes #330 
image


GiEnJul/ClientApp/src/components/admin/overview/ConnectButton.tsx
```diff
@@ -60,10 +60,6 @@ const ConnectButton: React.FC<ConnectButtonProps> = ({
       return <></>;
     }
   };
-  return (
-    <Grid container spacing={2} justifyContent="flex-end" alignItems="flex-end">
-      {displayButton()}
-    </Grid>
-  );
+  return <>{displayButton()}</>;
```
joffe97, line 63: Could the wrapper (`<>...</>`) be removed here?

GiEnJul/ClientApp/src/components/admin/overview/Statistics.tsx
```diff
@@ -0,0 +1,88 @@
...
+  let confirmedMatch: number = 0;
+  givers.map((giver) => {
+    if (giver.isSuggestedMatch) {
+      if (giver.hasConfirmedMatch) {
+        confirmedMatch++;
+      } else {
+        suggestedMatch++;
+      }
+    }
+  });
```
joffe97, line 25: This should have been done with a useEffect hook, with giver as second parameter. Now the givers list is iterated upon every render. The suggestedMatch and confirmedMatch should also be put into a state.
For this to work, the states must be set after the loop has finished. The javascript filter method could probably be used here?

joffe97 (dismissed):

GiEnJul/ClientApp/src/components/admin/overview/Statistics.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+type IStatistics = {
+  givers: GiverType[] | [];
+  recipients: RecipientType[] | [];
+};
+
+const Statistics: React.FC<IStatistics> = ({ givers, recipients }) => {
+  const totalGivers: number = givers.length;
+  const totalRecipints: number = recipients.length;
+  let suggestedMatch: number = 0;
```
Sommerli, line 15: Agree with Joachim, use filter to set these values
https://stackoverflow.com/a/47377879

GiEnJul/ClientApp/src/components/admin/overview/Statistics.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const recipientsWithoutGiver: number = totalRecipints - suggestedMatch;
+  const classes = useStyles();
+
+  return (
+    <>
+      <Grid container direction="column" justifyContent="flex-start" className={classes.statisticsContainer}>
+        <Grid item className={classes.infoContainer}>
+          <Typography  variant="h4">
+            {" "}
+            {totalRecipints.toString()} familier
```
Sommerli, line 36: toString calls are not needed on the numbers in rest of file

Sommerli (changes requested):
Keep in mind that admins might have a smaller screen(laptop etc). The legend on left is somewhat hidden on smaller screens. This could be solved by simply moving the Statistics component up a tiny bit.
image

Sommerli (approved):

joffe97 (approved):

PR #364, opened 2021-08-05
Bug when verifying connection 

elisabethnetland:
Removed exception-handling if user already exists


GiEnJul/Controllers/VerifyConnectionController.cs
```diff
@@ -47,11 +47,6 @@ public async Task VerifyConnection(string giverRowKey, string recipientRowKey, s
             recipient.FamilyMembers = await _personRepository.GetAllByRecipientId(recipient.RowKey);
             var giver = await _giverRepository.GetGiverAsync(partitionkey, giverRowKey);
 
-            if (await _connectionRepository.ConnectionExists(giver, recipient))
-            {
```
Sommerli, line 51: Keep this if, as its not necessary to continue if the connection is already made.

Put a return inside the if

Sommerli (changes requested):
Small change

TheseusDeus (dismissed):
Ok, as long as you do fix the return same as @Sommerli said

Sommerli (approved):

PR #391, opened 2021-08-06
Changed giver registration email

joffe97:
image

### DIFF
--- a/GiEnJul/Controllers/GiverController.cs
+++ b/GiEnJul/Controllers/GiverController.cs
@@ -52,7 +52,14 @@ namespace GiEnJul.Controllers
 
             var insertedAsDto = _mapper.Map<PostGiverResultDto>(await _giverRepository.InsertOrReplaceAsync(giver));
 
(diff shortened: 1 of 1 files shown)


GiEnJul/Controllers/GiverController.cs
```diff
@@ -52,7 +52,14 @@ public async Task<ActionResult<PostGiverResultDto>> PostAsync([FromBody] PostGiv
 
             var insertedAsDto = _mapper.Map<PostGiverResultDto>(await _giverRepository.InsertOrReplaceAsync(giver));
 
-            var messageContent = $"{insertedAsDto.FullName}, vi bekrefter å ha mottatt din registering for å gi en jul i {insertedAsDto.Location}";
+            var messageContent =
```
Sommerli, line 55: Should the text be equal to what they just saw on the website?

image

joffe97: Removed the `Tusen takk for innsatsen dere skal gjøre!` part. I think the rest is necessary.

elisabethnetland (approved):

PR #404, opened 2021-09-01
Inforce the use of azurite 3.13.1

### LINKED ISSUE #402: Query for "Foreslåtte koblinger" does not provide correct result
**This is a bug with Azurite 3.14.0**

**Describe the bug**
The query corresponding to the tab "Foreslåtte koblinger" in the Dev institution User admin page does not work.
The query results in each condition being surrounded by parentheses and, where the next condition is placed inside the previous one. This prevents the code from generating the correct query, which results in no "koblinger" being shown.

**To Reproduce**
1. Use azurite version `3.14.0`.
2. Log in to the admin page of dev institution user
3. Select the "Foreslåtte koblinger" tab.
4. Verify that no "koblinger" are shown

**Expected behavior**
The query should create the query without the excess parentheses, resulting in a valid query for the desired output.

HGodal:
Updates the readme to provide information that azurite version 3.13.1 has to be used to avoid some query-issues 3.14.0 has.

Closes #402

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,8 @@ Run to show current npm version:  -npm -v
 The website should now be accessible on localhost:3000
 
 ### DataBase
-We used *azurite* for emulating a Azure Table Storage locally.
-* Run - npm install -g azurite
+We used *azurite* 3.13.1 for emulating a Azure Table Storage locally.
+* Run - npm install -g azurite@3.13.1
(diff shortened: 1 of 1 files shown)


Sommerli (approved):

TTorgersen (approved):
Helpful fix.

PR #420, opened 2021-09-15
Fix410/change id placeholder fixes #410

### LINKED ISSUE #410: Change PID under family registration
Change placeholder text from PID 
image

to "PID eller annen ID dere benytter"
Something like that. Clarify that ID is not mandatory and should only be filled if the family has a id.

Sending a screenshot of the finished work to Karianne should be done, to verify that its done as she wished :)

TTorgersen:
Might need to reconsider the "onclick" function if it is not clear enough what the intention of it is. 
Could also just use static text to describe what PID is.

### DIFF
--- a/ClientApp/src/components/institution/InstitutionForm.tsx
+++ b/ClientApp/src/components/institution/InstitutionForm.tsx
@@ -6,6 +6,8 @@ import {
   Typography,
 } from "@material-ui/core";
 import { Alert } from "@material-ui/lab";
+import HelpOutlineIcon from '@material-ui/icons/HelpOutline';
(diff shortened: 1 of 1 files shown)


ClientApp/src/components/institution/InstitutionForm.tsx
```diff
@@ -114,7 +121,7 @@ const initValidFormState: ValidFormEntry = {
 };
 interface props {
   accessToken: string;
-}
+}
```
HGodal, line 124: 🤒

HGodal (approved):

SindreVatnaland (approved):
Looks good!

Sommerli (approved):

PR #455, opened 2021-09-27
Changed phone validation on institution to numeric

SindreVatnaland:


ClientApp/src/components/InputFields/Validators/Validators.tsx
```diff
@@ -8,10 +8,10 @@ export function isPhoneNumber(inputValue: string) {
   // Returns true if norwegian number or foreign number starting with +{countryCode}
   inputValue = inputValue.trim();
   return !!(
-    validator.isMobilePhone(inputValue, ["nb-NO", "nn-NO"]) ||
+    validator.isNumeric(inputValue) ||
     (inputValue &&
       inputValue.startsWith("+") &&
-      validator.isMobilePhone(inputValue))
+      validator.isNumeric(inputValue.replace("+", "")))
   );
```
HGodal, line 15: Could this cause some sort of problem if the giver enters i.e. "+47 123 45 678"? Maybe the + sign and the country code should be removed in that case?

PR #457, opened 2021-09-28
Updated react-api variable to PROD and some requested text changes 

Sommerli:
Updated env.production variable, as currently master calls the test-api and not prod-api

Also some small text changes, and a clickable _ofte stilte spørsmål_ text under contacts


TTorgersen (approved):

RonHaland (approved):
You can complete this one and I'll make a .env.development file and update the pipelines

Issue #452, opened 2021-09-27
Institutions not able to insert phone number


HGodal:
**Describe the bug**
Institutions cannot enter phone numbers starting with something else than 4 or 9, or have another length than 8.

**To Reproduce**
1. Log in as an institution and try to crate a family member.

**Expected behavior**
Phone number validation should be a bit more loose.  Drop the validation of the starting number and the length. Only verify that the input is a number.

Issue #456, opened 2021-09-27
Move information card together with the other cards on landing page


SindreVatnaland:
Replace "Elisabeth Ohlgren Tidemann" with "Vil du Gi en jul?" and place the card beside the other contacts.

![Issue](https://user-images.githubusercontent.com/69798990/134914897-5c61afde-8b5f-49c4-be3d-042d8eb34717.PNG)

PR #459, opened 2021-09-28
Cardadjustment/456

SindreVatnaland:
Moved the card about how to start "Gi en jul" in your municipality together with the contact cards.

Capture

#456


TTorgersen (dismissed):
Also fixes #452 ?
Thus PR #455 not neccesary

Issue #494, opened 2021-10-06
Suggested giver does not display unless all giversizes are selected


TTorgersen:
If there are few registered givers, and the givers does not have at least 1 familysize each, (< 3, 3-5, >5), the suggested family connections will not fetch the recommended givers.

This should be fixed to display either 3 different givers with different family sizes, or the three oldest registered givers that fit the suggested families.

PR #495, opened 2021-10-06
Fixed running into an exception if not all giversizes were present

Sommerli:
First would throw an error if there was no match, FirstOrDefault returns null if there is no match.

Closing #494


RonHaland:
Obsolete by #496?

PR #496, opened 2021-10-06
add tests for Admin/Suggestions/Giver

RonHaland:


GiEnJul.Test/ControllerTests/AdminControllerTest.cs
```diff
@@ -0,0 +1,115 @@
...
+                                              mockConnectionRepo.Object,
+                                              _log,
+                                              _mapper,
+                                              emailClientMock.Object,
+                                              settings);
+        }
+
+        public void Dispose()
+        {
+            mockEventRepo.VerifyNoOtherCalls();
```
Sommerli, line 52: Could add `mockGiverRepo.VerifyNoOtherCalls()` This will throw an exception if any calls to the GiverRepo is done, other than those expected. Expected being defined with `mockGiverRepo.Verify(x => x.MyMethod...`

PR #497, opened 2021-10-07
Added Dialog popup when registering a family

SindreVatnaland:
Dialog popup replaces the Success alert when successfully registering a family. The popup will also show the family Id as well as Reference Id given by the institution. 

Reference Id is conditionally rendered if given

Capture

### DIFF
--- a/ClientApp/src/components/institution/ConfirmationDialog.tsx
+++ b/ClientApp/src/components/institution/ConfirmationDialog.tsx
@@ -37,7 +37,7 @@ return (
                 { referenceId.length > 0 &&
                 <p>Din Id: <b>{referenceId}</b></p>
                 }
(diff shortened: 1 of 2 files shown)


ClientApp/src/components/institution/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,53 @@
...
+      >
+        <DialogTitle id="alert-dialog-title">
+          {"Familie Registrert"}
+        </DialogTitle>
+        <DialogContent>
+          <DialogContentText id="alert-dialog-description">
+                { referenceId.length > 0 &&
+                <p>Din Id: <b>{referenceId}</b></p>
+                }
+                <p>Gi en jul's samilienummer: <b>{familyId}</b></p>
```
Sommerli, line 40: samilienummer - typo

PR #500, opened 2021-10-07
Order connection suggestions

RonHaland:


RonHaland:
The list will first be ordered by date to get the givers that signed up first towards the top. They are then grouped into 3 groups based on their MaxReceivers and the first instance of each is selected. Then this list is sorted by maxReceivers so that the "small family" givers will appear at the top.

PR #501, opened 2021-10-13
Person comment fix

SindreVatnaland:
Its now more clear what is going on when adding a comment to a person in a family.
The clickable link will now change its text from "Legg til" to "Se/Endre" to visually identify that a comment is added.
Kommentar
Adding or updating a comment will send an alert to notify that the comment is added.
Alert
Closing the comment tab rather than saving will undo unsaved changes.


TTorgersen (approved):
Nice update.

Issue #414, opened 2021-09-15
Frontend - Use the location of the institute to avoid manual selection of location in Register Family


TTorgersen:
**Is your feature request related to a problem? Please describe.**
We would like the institution-admin page to know the location of the logged-in institution employee, thus eliminating the need to select a location manually: 
![image](https://user-images.githubusercontent.com/46501222/133421709-29eb9631-493e-4f77-8173-bc059f07a365.png)

**Describe the solution you'd like**
Use the information about the logged-in institution employee (per municipality) to remove unnecessary registration steps for the institution.

**Describe alternatives you've considered**
The institution user is registered with a location (municipality ID or name) and while rendering the register family page, the unnecessary information will not be displayed.

PR #514, opened 2021-10-13
Automatic location on institution page

SindreVatnaland:
Location on institution page is now removed and rather uses the location registered on the user to automatically select a location.
This is also the case for the Institution where eg. NAV or Barneværnet is read from its user rather than being statically added.

#414

### DIFF
--- a/ClientApp/src/hooks/useUser.tsx
+++ b/ClientApp/src/hooks/useUser.tsx
@@ -39,9 +39,9 @@ const reducer = (state: IState, action: IReducerAction) => {
     case ReducerActionType.setLocation:
       return { ...state, location: action.data ?? null };
     case ReducerActionType.setRole:
-      return { ...state, role: action.data ?? "Unspecified" };
+      return { ...state, role: action.data ?? null };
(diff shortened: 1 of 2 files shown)


ClientApp/src/hooks/useUser.tsx
```diff
@@ -36,6 +40,8 @@ const reducer = (state: IState, action: IReducerAction) => {
       return { ...state, location: action.data ?? null };
     case ReducerActionType.setRole:
       return { ...state, role: action.data ?? "Unspecified" };
+    case ReducerActionType.setInstitution:
+      return { ...state, institution: action.data ?? "Unspecified" };
```
Sommerli, line 44: Should this be "Unspecified" or null?

Sommerli (commented):
Not all users will have the institution metadata set(admins), does these changes interfere with other useUser calls?

ClientApp/src/hooks/useUser.tsx
SindreVatnaland: User without set metadata will not be affected

PR #518, opened 2021-10-20
379/endpoint removing connections

HGodal:
Open for suggestions/improvements

The new endpoint `admin/connection?location=...&rowKey=...` takes location and rowkey of a giver as input, and deletes the corresponding connection. This includes both completed connections and connections waiting to be confirmed.

### DIFF
--- a/GiEnJul/Controllers/AdminController.cs
+++ b/GiEnJul/Controllers/AdminController.cs
@@ -132,6 +132,68 @@ namespace GiEnJul.Controllers
             return Ok();
         }
 
+        [HttpDelete("Connection")]
+        // [Authorize(Policy = "DeleteConnection")]
+        public async Task<ActionResult> DeleteConnectionAsync(string location, string rowKey)
+        {
+            Giver giver;
+            Recipient recipient;
+
+            try
+            {
+                giver = await _giverRepository.GetGiverAsync(location, rowKey);
+                recipient = await _recipientRepository.GetRecipientAsync(location, giver.MatchedRecipient);
+
+            }
+            catch (NullReferenceException)
+            {
+                return NotFound();
+            }
+
+            catch (ArgumentNullException)
+            {
+                return NoContent();
+            }
+
+            var originalMatchedRecipient = giver.MatchedRecipient;
+            var originalMatchedGiver = recipient.MatchedGiver;
+
+            giver.HasConfirmedMatch = false;
+            giver.IsSuggestedMatch = false;
+            giver.MatchedRecipient = null;
+
+            recipient.HasConfirmedMatch = false;
+            recipient.IsSuggestedMatch = false;
+            recipient.MatchedGiver = null;
+
+            try
+            {
+                if (await _connectionRepository.ConnectionExists(giver, recipient))
+                {
+                    await _connectionRepository.DeleteConnectionAsync(location, recipient.RowKey + "_" + giver.RowKey);
+                }
+                await _giverRepository.InsertOrReplaceAsync(giver);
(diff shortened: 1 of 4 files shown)


GiEnJul/Controllers/AdminController.cs
```diff
@@ -132,6 +132,58 @@ public async Task<ActionResult> PutWishAsync(string person_rowkey, [FromBody] st
...
 
+        [HttpDelete("Connection")]
+        [Authorize(Policy = "DeleteConnection")]
+        public async Task<ActionResult> DeleteConnectionAsync(string location, string rowKey)
+        {
+            var giver = await _giverRepository.GetGiverAsync(location, rowKey);
+
+            if (giver?.MatchedRecipient is null)
+            {
+                return NoContent();
```
Sommerli, line 143: I belive `return NotFound();` would be more fitting here

https://docs.microsoft.com/en-us/aspnet/core/web-api/action-return-types?view=aspnetcore-5.0#synchronous-action-1

GiEnJul/Controllers/AdminController.cs
```diff
@@ -132,6 +132,58 @@ public async Task<ActionResult> PutWishAsync(string person_rowkey, [FromBody] st
...
+            }
+
+            var recipient = await _recipientRepository.GetRecipientAsync(location, giver.MatchedRecipient);
+
```
Sommerli, line 147: Could also do a null check on recipient here

GiEnJul/Controllers/AdminController.cs
```diff
@@ -132,6 +132,58 @@ public async Task<ActionResult> PutWishAsync(string person_rowkey, [FromBody] st
...
+            {
+                await _giverRepository.InsertOrReplaceAsync(giver);
+                await _recipientRepository.InsertOrReplaceAsync(recipient);
+
+                if (await _connectionRepository.ConnectionExists(giver, recipient))
+                {
+                    await _connectionRepository.DeleteConnectionAsync(location, recipient.RowKey + "_" + giver.RowKey);
+                }
+            }
+            catch (NullReferenceException)
```
Sommerli, line 169: The operations in the try block could throw other exceptions than NullReferenceException. Should another exception be thrown, the tables might get faulty data as not all operations are completed.  I suggest you change from NullReferenceException to Exception e, and do a _log.Error call.

GiEnJul/Controllers/AdminController.cs
HGodal: Will do

GiEnJul/Controllers/AdminController.cs
HGodal: 👍🏼

TTorgersen (approved):

GiEnJul/Controllers/AdminController.cs
```diff
@@ -132,6 +132,56 @@ public async Task<ActionResult> PutWishAsync(string person_rowkey, [FromBody] st
...
+            recipient.HasConfirmedMatch = false;
+            recipient.IsSuggestedMatch = false;
+            recipient.MatchedGiver = null;
+
+            try
+            {
+                await _giverRepository.InsertOrReplaceAsync(giver);
+                await _recipientRepository.InsertOrReplaceAsync(recipient);
+
+                if (await _connectionRepository.ConnectionExists(giver, recipient))
```
Sommerli: This if and contents should be put outside the try catch. Should the if be false, the giver and recipient while no connection is updated.

The if and its contents should probably be put before the try catch.

Sommerli (approved):

PR #520, opened 2021-10-27
Edit registered families/516

SindreVatnaland:
When on SuggestedConnections you can now edit a selected family and family members information.


GiEnJul/Controllers/AdminController.cs
```diff
@@ -240,6 +240,25 @@ public async Task<ActionResult> DeleteRecipientAsync([FromBody] DeleteRecipientD
             return Ok();
         }
 
+        [HttpPut("recipient")]
+        [Authorize(Policy = "UpdateRecipient")]
+        public async Task<ActionResult> PutRecipientAsync([FromBody] PutRecipientDto recipientDto)
+        {
+            var recipientToPut = _mapper.Map<Recipient>(recipientDto);
+            var recipientToSet = await _recipientRepository.GetRecipientAsync(recipientDto.PartitionKey, recipientDto.RowKey);
```
Sommerli, line 248: Naming suggestion:
recipientToPut -> recipientNew, updatedRecipient, or keep as is
recipientToSet -> recipientOld , recipientToUpdate

GiEnJul/Controllers/AdminController.cs
```diff
@@ -240,6 +240,25 @@ public async Task<ActionResult> DeleteRecipientAsync([FromBody] DeleteRecipientD
...
+            foreach (var prop in recipientToPut.GetType().GetProperties())
+            {
+                var propValuePut = prop.GetValue(recipientToPut);
```
Sommerli, line 251: propValuePut cease to exits after } , which means it could contain a simpler/generic name, i suggest `value` or `newValue`

GiEnJul/Controllers/AdminController.cs
```diff
@@ -240,6 +240,25 @@ public async Task<ActionResult> DeleteRecipientAsync([FromBody] DeleteRecipientD
...
+                if (propValuePut != null) prop.SetValue(recipientToSet, propValuePut);
+            }
+            await _recipientRepository.InsertOrReplaceAsync(recipientToSet);
+            foreach (var person in recipientToSet.FamilyMembers)
```
Sommerli, line 255: This operation should be within a try catch.
Given a recipient is updated and the code fails during updating the person-list, all operations should be undone to not leave faulty data.

See 
https://github.com/capgemini-stavanger/gi-en-jul/blob/b75a8ad8b69ffc6987a8acb8eb712b59815ff29a/GiEnJul/Controllers/RecipientController.cs#L53

ClientApp/src/components/admin/suggestedConnections/RecipientTable.tsx
```diff
@@ -10,34 +10,39 @@ import TableRow from "@material-ui/core/TableRow";
...
 import KeyboardArrowUpIcon from "@material-ui/icons/KeyboardArrowUp";
+import EditIcon from '@material-ui/icons/Edit';
 import { RecipientType } from "./Types";
 import { getGender } from "../../../common/functions/GetGender";
 import { capitalize } from "@material-ui/core";
 import { Group } from "@material-ui/icons";
 import useStyles from "../common/Styles";
 
-type RecipientRowProps = {
+type IRecipientRowProps = {
```
Sommerli, line 20: I is used for identifying interfaces

GiEnJul/Controllers/AdminController.cs
SindreVatnaland: recipientToPut is now recipientUpdated
recipientToSet is now recipientToUpdate

PR #532, opened 2021-11-03
Added necessary files to update Mail #1 to newest version

TTorgersen:


TTorgersen:
I tested locally with registering a giver that the system works and that the dynamic Email formatting was correct.

PR #536, opened 2021-11-06
EditFamilies from Overview Page

SindreVatnaland:
Families can now be edited from the admin overview page

### DIFF
--- a/ClientApp/src/components/admin/common/RecipientTable.tsx
+++ b/ClientApp/src/components/admin/common/RecipientTable.tsx
@@ -5,6 +5,7 @@ import {
   Container,
   Divider,
   Typography,
+  IconButton,
 } from "@material-ui/core";
 import {
   ExpandMore,
@@ -17,15 +18,19 @@ import * as React from "react";
 import Gender from "../../../common/enums/Gender";
 import { RecipientType } from "../overview/Types";
 import useStyles from "./Styles";
+import EditIcon from '@material-ui/icons/Edit';
+import { useState } from "react";
 
 type Props = {
   data: RecipientType[] | [];
   handleRecipientChange: (newRecipient: RecipientType) => void;
+  openDialog: () => void;
 };
 
 const DatatableRecipient: React.FC<Props> = ({
   data,
   handleRecipientChange,
+  openDialog,
 }) => {
   const classes = useStyles();
 
@@ -34,6 +39,8 @@ const DatatableRecipient: React.FC<Props> = ({
     setExpanded(isExpanded ? recipient : false);
     handleRecipientChange(recipient);
   };
+
+  const [selected, setSelected] = useState(-1);
 
   const formatFamily = (input: Number) => {
     if (input < 3) {
@@ -72,13 +79,15 @@ const DatatableRecipient: React.FC<Props> = ({
 
   return (
     <Container>
-      {data.map((recipient) => (
-        <Accordion expanded={expanded===recipient}
-          onChange={handleChange(recipient)}
+      {data.map((recipient, index) => (
+        <Accordion
+          expanded={selected == index}
           key={recipient.rowKey}
           className={classes.accordionContainer}
+          onClick={() => { index != selected ? setSelected(index) : setSelected(-1)}}
         >
           <AccordionSummary
+            onClick={() => handleRecipientChange(recipient)}
             expandIcon={<ExpandMore />}
             aria-controls="panel1bh-content"
             id="panel1bh-header"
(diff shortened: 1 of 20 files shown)


ClientApp/src/components/admin/common/RecipientTable.tsx
```diff
@@ -72,13 +79,15 @@ const DatatableRecipient: React.FC<Props> = ({
...
     <Container>
-      {data.map((recipient) => (
-        <Accordion expanded={expanded===recipient}
-          onChange={handleChange(recipient)}
+      {data.map((recipient, index) => (
+        <Accordion
+          expanded={selected == index}
           key={recipient.rowKey}
           className={classes.accordionContainer}
+          onClick={() => { index != selected ? setSelected(index) : setSelected(-1)}}
```
Sommerli, line 87: Should onClick use handleChange? 

handleChange function is unused after this changes

ClientApp/src/components/admin/overview/OverviewMacro.tsx
```diff
@@ -78,14 +83,28 @@ const OverviewMacro: React.FC<IOverviewMacro> = ({ accessToken, location }) => {
...
+              dinner: newRecipient.dinner,
+              dessert: newRecipient.dessert,
+              note: newRecipient.note,
+              contactFullName: newRecipient.contactFullName,
+              contactEmail: newRecipient.contactEmail,
+              contactPhoneNumber: newRecipient.contactPhoneNumber,
+              institution: newRecipient.institution,
+              referenceId: newRecipient.referenceId,
+              familyMembers: newRecipient.familyMembers as Types.PersonType[],
+            } as Types.RecipientType
```
Sommerli, line 107: Maybe this function could easily map the properties?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

The usage of two classes with the same name is confusing. Why are they seperated and should they be seperate? 

Might be an idea to remove usage of one in favour of the other, or create a subclass

SindreVatnaland: I were no sure if that function were working due to the recipient objects originates from different types, but in fact it does!

I got it working writing it like this:
```cs
Object.assign({} as Types.RecipientType, recipient);
```

ClientApp/src/components/admin/common/RecipientTable.tsx
SindreVatnaland: Created a handleSelectedAccordion() function!

ClientApp/src/components/institution/FamilyDialog.tsx
```diff
@@ -10,9 +10,11 @@
  import Recipient from "../admin/overview/Recipient";
  import ApiService from "../../common/functions/apiServiceClass";
  import { useEffect, useState, useCallback} from "react";
+ import * as Types from "../admin/suggestedConnections/Types";
```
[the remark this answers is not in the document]
SindreVatnaland: Moved this to ClientApp\src\common\components\Types.tsx

GiEnJul/Controllers/AdminController.cs
```diff
@@ -288,6 +288,33 @@ public async Task<ActionResult> DeleteRecipientAsync([FromBody] DeleteRecipientD
...
+            }
+            catch (Exception ex)
+            {
+                throw ex;
+            }
+            await _recipientRepository.InsertOrReplaceAsync(recipientToUpdate);
+            foreach (var person in recipientToUpdate.FamilyMembers)
+            {
+                await _personRepository.InsertOrReplaceAsync(person);
+            }
```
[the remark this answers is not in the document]
SindreVatnaland: Like this?:
```cs
public async Task<ActionResult> PutRecipientAsync([FromBody] PutRecipientDto recipientDto)
        {
            var recipientUpdated = _mapper.Map<Recipient>(recipientDto);
            var recipientToUpdate = await _recipientRepository.GetRecipientAsync(recipientDto.PartitionKey, recipientDto.RowKey);

            try
            {
                foreach (var prop in recipientUpdated.GetType().GetProperties())
                {
                    var value = prop.GetValue(recipientUpdated);
                    if (value != null) prop.SetValue(recipientToUpdate, value);
                }


                await _recipientRepository.InsertOrReplaceAsync(recipientToUpdate);
                foreach (var person in recipientToUpdate.FamilyMembers)
                {
                    await _personRepository.InsertOrReplaceAsync(person);
                }
            }
            catch (Exception ex)
            {
                throw ex;
            }
            return Ok();
        }
```

ClientApp/src/components/institution/FamilyDialog.tsx
```diff
@@ -70,7 +71,7 @@ const FamilyDialog: React.FC<IFamilyDialog> = ({ open, accessToken,institution,
                      <DialogContentText id="alert-dialog-description">
                          <p>Liste over familier</p>
                      </DialogContentText>
-                     <Recipient data={recipientData} handleRecipientChange={handleRecipientChange}/>
+                     <Recipient data={recipientData} accessToken={""} refreshRecipients={() => ""}/>
```
Sommerli, line 74: Should this include accessToken if institutions should want to change families?

Sommerli (approved):

PR #632, opened 2022-06-23
validating refernceId backend

### LINKED ISSUE #631: Add family form - validate refernce ID - Backend
Backend validating of referenceId

En-Lun:
Validating referenceId backend.

added check in controller and a function in the repository

close #631

### DIFF
--- a/GiEnJul/Repositories/RecipientRepository.cs
+++ b/GiEnJul/Repositories/RecipientRepository.cs
@@ -49,9 +49,11 @@ namespace GiEnJul.Repositories
         {
             var query = $"ReferenceId eq '{referenceId}";
             var exists = await GetAllByQueryAsync(query);
-            if (_mapper.Map<List<Models.Recipient>>(exists).Count > 0) return true;
+            if (_mapper.Map<List<Models.Recipient>>(exists).Count > 0)
+                return true;
            
-            else return false;
+            else
+                return false;
             
         }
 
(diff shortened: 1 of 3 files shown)


hildmi (approved):

GiEnJul/Repositories/RecipientRepository.cs
```diff
@@ -44,6 +45,16 @@ public async Task<List<Models.Recipient>> GetUnmatchedRecipientsAsync(string loc
             return _mapper.Map<List<Models.Recipient>>(recipients);
         }
 
+        public async Task<bool> RecipientDoesExist(string referenceId)
+        {
+            var query = $"ReferenceId eq '{referenceId}";
+            var exists = await GetAllByQueryAsync(query);
+            if (_mapper.Map<List<Models.Recipient>>(exists).Count > 0) return true;
```
RonHaland, line 52:
```suggestion
            if (_mapper.Map<List<Models.Recipient>>(exists).Any()) 
                 return true;
```
But I think this change also requires "using linq" 🤷‍♂️
Anyways, please move the return down

GiEnJul/Repositories/RecipientRepository.cs
```diff
@@ -44,6 +45,16 @@ public async Task<List<Models.Recipient>> GetUnmatchedRecipientsAsync(string loc
...
+           
+            else return false;
```
RonHaland, line 54:
```suggestion
            else 
                 return false;
```
Move these down

RonHaland (commented):
Just some visual feedback 😁

MagnusVG (dismissed):

GiEnJul/Repositories/RecipientRepository.cs
En-Lun: fixed

GiEnJul/Repositories/RecipientRepository.cs
En-Lun: fixed

MagnusVG (approved):

PR #762, opened 2022-07-08
Delete entry

### LINKED ISSUE #756: Cms Table: Make new endpoint for deleting entries in Table
1. Make endpoint in Cms controller: /delete with parameter : PartitionKey & RowKey
2. Make method in repository to delete entry in the Cms Table

hildmi:
closes #756


ClientApp/src/components/superadmin/kommune/KommuneInformation.tsx
```diff
@@ -37,6 +37,23 @@ const KommuneInformation: React.FC<IKommuneInformation> = ({ accessToken, locati
   };
   useEffect(fetchInformation, [location]);
 
+  const deleteLocation = () => {
+    apiservice
+      .post("Cms/deleteSingle", {
+        ContentType: "Kommune",
+        Index: location,
+        Question: "",
+        Info: "",
```
RonHaland, line 46: Question and Info can be removed here, no reason to send empty strings to back end as long as the corresponding properties are nullable in the dto class.

RonHaland (approved):

PR #813, opened 2022-07-15
Managment Api created, and route for user creating added.

YohannesDK:


GiEnJul/Dtos/CreateUserDto.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace GiEnJul.Dtos
+{
+    public class CreateUserDto
+    {
+        public string Email { get; set; }
+        public string Password { get; set; }
+        public string Location { get; set; }
+        public string Role { get; set; }
+        public string? Institution { get; set; }
```
RonHaland, line 9:
```suggestion
        public string Institution { get; set; }
```
We don't have the "nullable" feature enabled in this project so all strings are nullable

RonHaland (approved):
I suggest removing the '?' to prevent github actions from complaining

GiEnJul/Dtos/CreateUserDto.cs
YohannesDK: okey 👍

RonHaland (approved):

En-Lun (approved):

MagnusVG (approved):

hildmi (approved):

PR #816, opened 2022-07-16
add endpoint for creating excel table

RonHaland:
Part 1 of #787

### DIFF
--- a/GiEnJul/Controllers/RecipientController.cs
+++ b/GiEnJul/Controllers/RecipientController.cs
@@ -1,14 +1,18 @@
-﻿using GiEnJul.Dtos;
-using GiEnJul.Models;
+﻿using AutoMapper;
+using ClosedXML.Extensions;
 using GiEnJul.Auth;
+using GiEnJul.Dtos;
+using GiEnJul.Models;
+using GiEnJul.Repositories;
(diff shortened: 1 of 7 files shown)


En-Lun (approved):

MagnusVG (approved):

GiEnJul/Controllers/RecipientController.cs
```diff
@@ -82,13 +86,39 @@ public async Task<ActionResult<(string FamilyId,string ReferenceId)>> PostAsync(
...
+            var persons = await _personRepository.GetAllByRecipientIds(recipientIds);
+
+            recipients.ForEach(r => r.FamilyMembers = persons.Where(p => p.RecipientId == r.RecipientId).ToList());
+
             return recipients;
         }
+
+        [HttpGet("excel")]
+        [Authorize(Policy = Policy.ReadRecipient)]
+        public async Task<FileStreamResult> GetExcelListForInstitutionAsync([FromQuery] string institution)
```
MagnusVG, line 100: Can we be using the user metadata instead of sending in the institution?

RonHaland: Probably. Will try it out when I do the rest of #787

PR #900, opened 2022-07-29
added appsettings.json to gitignore

havardMoe:
This will prevent pushing secrets to git


RonHaland (changes requested):
Is there an appsettings.json for frontend? Secrets shouldn't really be in the frontend.


If this is supposed to be for backend, instead add secrets locally by right clicking the project file in solution explorer and click "project secrets" 
Just add a placeholder value in appsettings.json

PR #936, opened 2022-08-02
Email bug

YohannesDK:


GiEnJul/Controllers/EmailController.cs
```diff
@@ -78,7 +79,7 @@ public async Task<ActionResult> SendMail(PostEmailFromUserDto email)
         }
 
         [HttpPost("sendFromUser")]
-        [Authorize(Policy = Policy.PostEmail)]
+        //[Authorize(Policy = Policy.PostEmail)]
```
RonHaland, line 82: Is this supposed to be commented out?

RonHaland (approved):
Approved with one question

GiEnJul/Controllers/EmailController.cs
YohannesDK: No :/ i will fix

RonHaland (approved):

PR #1086, opened 2026-02-01
Chore/update deps

RonHaland:
Before this PR is merged to dev:
1. Update DEPLOY_SAS_URL and DEPLOY_SAS_URL_PROD [secrets](https://github.com/capgemini-stavanger/gi-en-jul/settings/secrets/actions) with the instructions from [the wiki here](https://github.com/capgemini-stavanger/gi-en-jul/wiki/Azure#cd). The SAS URL must be for the storageAccount that hosts the static website.
2. update the runtime (dotnet version) config on the appservice in azure. The DEV appservice first, and then PROD before merging dev to master.

Also includes 2 fixes in the API that gave errors instead of empty responses.

### DIFF
--- a/.github/workflows/CI_backend.yml
+++ b/.github/workflows/CI_backend.yml
@@ -22,10 +22,10 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Set up .NET 10
-        uses: actions/setup-dotnet@v3
+        uses: actions/setup-dotnet@v5
         with:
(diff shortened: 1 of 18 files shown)


andreas-nesheim (approved):
Updated `DEPLOY_SAS_URL` and `DEPLOY_SAS_URL_PROD` with new SAS URL's. Used the values from the "Blob service SAS URL" - not sure if this is the correct one since there were four different ones (Blob service, file service, queue service and table service URLs). We'll find out when this deploys to dev 😄

Also updated the dev app service's .NET version to .NET 10

